PR #29, opened 2026-06-21
Dynamic negentropy targeting from the NIP-66 catalog

erskingardner:
## Summary

Completes the core **P1 collection loop**: negentropy reconciliation now follows *discovered* NIP-77 support from the NIP-66 catalog instead of a static, hand-maintained list.

- **`RelayManager::catalog_negentropy_targets(limit)`** — URL relays in the catalog that advertise NIP-77, payment-required relays excluded, ordered best-RTT-first, capped.
- **Startup augmentation** — the configured/default negentropy targets are merged (deduped, capped by `--negentropy-max-targets`, default 16) with catalog NIP-77 relays. Because the catalog only holds relays that *advertise* NIP-77, non-supporting relays (e.g. Primal) are naturally excluded — no special-casing. `--no-negentropy-catalog-targets` disables it.
- **`ingest_negentropy_targets` gauge** — completes the catalog-coverage picture: `known` (catalog) / `reconciled` (this) / `connected` (`relay_manager_active_relays`).

It **augments** rather than replaces the configured list, so there's no regression if the catalog is empty. The catalog persists in `relay-stats.db`, so it's populated across restarts (~70 NIP-77 relays in prod today).

## Design notes / follow-ups
- **Startup-time selection**, not per-cycle: targets are chosen once when the syncer is built (the syncer holds a fixed relay list). New NIP-77 relays discovered mid-run are picked up on the next restart. Per-cycle refresh is a clean follow-up.
- `supported_nips` matching uses the comma-wrapped `LIKE` trick on the comma-joined column — fine at this scale; a child table / JSON would be cleaner if NIP-match queries proliferate.

## Test plan
- `just precommit` green (fmt, clippy `-D warnings`, tests incl. a new `catalog_negentropy_targets` test covering NIP-77 selection, payment/non-URL exclusion, and RTT ordering).
- After deploy: `ingest_negentropy_targets` jumps from the small default to default+catalog (capped), and negentropy sync activity should broaden across the NIP-77 set.

## Plan status
P1 collection: reference-coverage ✅ · NIP-66 catalog ✅ · catalog-coverage metrics ✅ · **dynamic negentropy targeting ✅** → only **windowed REQ backfill** remains.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/pensieve-ingest/src/main.rs
+++ b/crates/pensieve-ingest/src/main.rs
@@ -263,6 +263,15 @@ struct Args {
     #[arg(long, default_value = "16")]
     negentropy_max_targets: usize,
 
+    /// Minimum distinct monitors that must report NIP-77 for a catalog relay to be
+    /// used as a negentropy target (quorum; guards against a single bad monitor).
+    #[arg(long, default_value = "2")]
+    negentropy_target_min_monitors: usize,
+
+    /// Max age (days) of a monitor observation for it to count toward target selection.
+    #[arg(long, default_value = "7")]
+    negentropy_target_max_age_days: u64,
+
     // ─────────────────────────────────────────────────────────────────────────
     // Coverage instrumentation
     // ─────────────────────────────────────────────────────────────────────────
@@ -508,7 +517,7 @@ async fn main() -> Result<()> {
     // Initialize negentropy sync (if enabled)
     let negentropy_syncer = if args.negentropy {
         // Load trusted relays for negentropy sync
-        let mut negentropy_relays = if let Some(ref relays) = args.negentropy_relays {
+        let negentropy_relays = if let Some(ref relays) = args.negentropy_relays {
             tracing::info!(
                 relay_count = relays.len(),
                 "using negentropy relays from CLI arguments"
@@ -544,35 +553,6 @@ async fn main() -> Result<()> {
             tracing::info!("Using default negentropy relays");
             NegentropySyncConfig::default().relays
         };
-
-        // Augment with NIP-77 relays from the NIP-66 catalog (deduped, capped).
-        // Catalog relays advertise NIP-77, so non-supporting relays (e.g. Primal)
-        // are naturally excluded. The catalog persists in relay-stats.db, so it is
-        // populated across restarts.
-        if !args.no_negentropy_catalog_targets {
-            match relay_manager.catalog_negentropy_targets(args.negentropy_max_targets) {
-                Ok(catalog_targets) => {
-                    let before = negentropy_relays.len();
-                    for t in catalog_targets {
-                        if !negentropy_relays.contains(&t) {
-                            negentropy_relays.push(t);
-                        }
-                    }
-                    let added = negentropy_relays.len() - before;
-                    if added > 0 {
-                        tracing::info!(
-                            added,
-                            total = negentropy_relays.len(),
-                            "added NIP-77 negentropy targets from catalog"
-                        );
-                    }
-                }
-                Err(e) => tracing::warn!(
-                    error = %compact_error(&e),
-                    "failed to load catalog negentropy targets"
-                ),
-            }
-        }
 
         // Open sync state database
         let sync_state = Arc::new(SyncStateDb::open(&args.negentropy_db_path).with_context(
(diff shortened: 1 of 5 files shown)


crates/pensieve-ingest/src/relay/manager.rs
```diff
@@ -270,6 +270,34 @@ impl RelayManager {
...
+    /// URL relays from the catalog that advertise NIP-77 (negentropy), best RTT
+    /// first, capped at `limit`. Relays requiring payment are excluded. Used to
+    /// pick negentropy reconciliation targets dynamically — relays that don't
+    /// support NIP-77 (e.g. Primal) are naturally absent.
+    pub fn catalog_negentropy_targets(&self, limit: usize) -> Result<Vec<String>> {
+        let conn = self.conn.lock();
+
+        let mut stmt = conn
+            .prepare(
+                "SELECT relay_id FROM relay_catalog
```
erskingardner, line 297: Blocking-ish: this selects any relay if any archived kind-30166 event says it has NIP-77. `record_relay_discovery` records 30166s from the normal firehose, not just a trusted monitor set, and NIP-66 explicitly warns that a monitor may be erroneous/malicious and clients should not trust a single source. That means one forged/misconfigured monitor can inject sync targets, up to `--negentropy-max-targets`, and have the negentropy task connect to them every cycle. I think this needs a trust gate before becoming an active reconciliation target: trusted monitor allowlist, quorum across multiple monitors, recency, and/or WoT filtering.

crates/pensieve-ingest/src/relay/manager.rs
```diff
@@ -270,6 +270,34 @@ impl RelayManager {
...
+                 WHERE relay_id_kind = 'url'
+                   AND (',' || supported_nips || ',') LIKE '%,77,%'
+                   AND (',' || requirements || ',') NOT LIKE '%,payment,%'
```
erskingardner: The `payment` exclusion is applied before `GROUP BY`, so it does not exclude a relay when one monitor row reports `R payment` and another row omits it or reports stale `!payment`; the payment row is simply filtered out and the relay still survives the group. If the selection contract is that relays requiring payment are excluded, aggregate first and use `HAVING` to reject an any/current/quorum payment requirement, or require an affirmative `!payment` signal from the same trusted/quorum set used for NIP-77 support.

crates/pensieve-ingest/src/main.rs
```diff
@@ -533,6 +545,35 @@ async fn main() -> Result<()> {
             NegentropySyncConfig::default().relays
         };
 
+        // Augment with NIP-77 relays from the NIP-66 catalog (deduped, capped).
+        // Catalog relays advertise NIP-77, so non-supporting relays (e.g. Primal)
+        // are naturally excluded. The catalog persists in relay-stats.db, so it is
+        // populated across restarts.
+        if !args.no_negentropy_catalog_targets {
+            match relay_manager.catalog_negentropy_targets(args.negentropy_max_targets) {
```
erskingardner: This is evaluated only once during startup, before the relay source begins recording any new 30166s in this process. A fresh or empty catalog, or a catalog that changes while the daemon runs, leaves the negentropy target list stuck until restart, even though the PR/docs describe this as dynamic and say it follows discovered support. If the goal is dynamic targeting, the syncer probably needs to refresh `catalog_negentropy_targets()` per sync cycle/interval, or receive a rebuilt target list when the catalog changes.

erskingardner (commented):
Adversarial pass: I found a few behavioral concerns around using the NIP-66 catalog as an active negentropy target source. The targeted selector test passes locally; these are production-semantics issues rather than compile failures.

erskingardner:
Thanks for the adversarial pass — all three addressed in 31ab83b:

**1. Trust gate (single-monitor injection).** `catalog_negentropy_targets` now requires NIP-77 to be asserted by **≥ `min_monitors` distinct monitors** (default 2, `--negentropy-target-min-monitors`) observed within **`max_age`** (default 7d, `--negentropy-target-max-age-days`). Rows are one-per-`(relay_id, monitor_pubkey)`, so `HAVING SUM(<nip77>) >= ?` is a genuine distinct-monitor quorum — one forged/misconfigured monitor can no longer inject a target, and recency keeps stale assertions from counting.

**2. Payment exclusion moved past `GROUP BY`.** It's now an aggregate: `HAVING SUM(CASE WHEN requirements ~ payment ...) = 0`, so a relay is rejected if **any** recent monitor reports `payment`, rather than just filtering that one row out. The full gate (NIP-77 quorum + no-payment + recency) runs on the aggregated group.

**3. Startup-only evaluation → per-cycle.** The syncer takes a `TargetProvider` (`NegentropySyncer::with_target_provider`) that's **re-queried every `sync_once`**, merged+deduped with the configured base relays — so the target set follows catalog growth (and a cold/empty catalog) without a restart. `ingest_negentropy_targets` is now emitted per cycle to reflect the live count.

Test rewritten to cover the quorum threshold, any-monitor-payment rejection, and recency. `just precommit` green.

One scoping note: this implements **quorum + recency** (NIP-66's "query multiple monitors" guidance). A curated trusted-monitor allowlist / WoT weighting would be a further hardening — easy to add as `--negentropy-trusted-monitors` in a follow-up if you'd like it.

PR #31, opened 2026-07-26
Add canonical Parquet lake migration and archive campaign

erskingardner:
## Summary

- define and accept the canonical V1 Nostr Parquet archive schema
- add the deterministic writer, strict validator, interoperability fixtures, and notepack conversion CLI
- add resumable SQLite-journaled local/S3 publication with immutable keys, conditional writes, quarantine, and atomic activation
- add the opt-in live sealed-segment Parquet shadow path
- add a bounded-disk historical campaign worker, systemd unit, durable receipts, cleanup, duplicate-representation filtering, and empty-segment no-op handling
- update the migration plan and writer benchmark with the real Hetzner/object-storage results
- stage the ClickHouse `events_by_time` projection removal and rollback outside the active migration glob
- restrict archive synchronization to sealed `.notepack` and `.notepack.gz` files

## Operational boundaries

This PR does not activate the historical campaign, deploy ClickHouse on the conversion VM, or enable live Parquet shadowing.

- `pensieve-parquet-campaign.service` is installed/enabled only by an explicit operator action.
- live shadow publication remains disabled unless a Parquet shadow target flag is supplied to `pensieve-ingest`.
- the prepared ClickHouse projection removal lives under `docs/migrations/prepared/`, so `just ch-migrate-all` cannot execute it.
- the existing Storage Box source archive remains read-only during historical conversion.

## Validation

- `just precommit`
- `just parquet-interop` — PyArrow 25.0.0 and DuckDB 1.5.5
- real Hetzner/object-storage canary: 389,697 input frames, 389,695 canonical rows, 2 quarantined rejects, independent download/checksum/strict validation
- same-journal retry completed as a fast no-op; a fresh journal regenerated identical bytes and converged on the existing immutable keys
- bounded systemd canary: 585,043 canonical rows, zero rejects, independent download and strict validation, durable receipt, staging/input cleanup
- real zero-byte Notepack and empty-gzip fixtures: durable zero-object publication, per-source receipts, and zero work on the second pass
- rclone sealed-name fixture: `.notepack.open` and `.notepack.gz.open` excluded


ops/systemd/pensieve-parquet-campaign.service
```diff
@@ -0,0 +1,37 @@
...
+ConditionPathExists=/etc/pensieve/parquet-object-storage.env
+ConditionPathExists=/etc/pensieve/parquet-campaign.env
+
+[Service]
+Type=simple
+User=pensieve
+Group=pensieve
+EnvironmentFile=/etc/pensieve/parquet-object-storage.env
+EnvironmentFile=/etc/pensieve/parquet-campaign.env
+ExecStart=/srv/pensieve/repo/ops/scripts/run-parquet-archive-campaign.sh
```
erskingardner: ExecStart hardcodes /srv/pensieve/repo/..., but every other unit in this repo (and ops/RUNBOOK.md:3: "The box runs at /home/pensieve/pensieve") uses /home/pensieve/pensieve. /srv/pensieve doesn't appear anywhere else in the tree. Installing this unit per the standard RUNBOOK flow will make it fail immediately (path doesn't exist), and with Restart=on-failure/RestartSec=30s and no StartLimitBurst, it will restart-loop forever. Same issue in ops/scripts/run-parquet-archive-campaign.sh:80 and ops/parquet-campaign.env.example:11.

crates/pensieve-lake/src/bin/pensieve-parquet-campaign.rs
```diff
@@ -0,0 +1,183 @@
...
+                if path.is_file() && is_notepack_segment(&path) {
+                    result.push(path);
+                }
+            }
+        } else {
+            result.push(input.clone());
+        }
+    }
+    result.sort();
+    result.dedup();
```
erskingardner, line 175: `result.dedup()` only drops *consecutive, byte-identical* PathBuf entries after a plain sort — it does not detect that `segment-1.notepack` and `segment-1.notepack.gz` represent the same logical segment. Work-unit identity is also purely `notepack-sha256-<source file bytes>` (work_unit.rs:104), so the two encodings hash differently and get converted/published as two separate immutable object sets containing the same events. The PR description's "duplicate-representation filtering" is actually implemented only as an awk heuristic in `ops/scripts/run-parquet-archive-campaign.sh` (prefers .gz, skips the paired .notepack) — running this binary directly in directory mode (as documented/intended for ad-hoc/non-systemd use) bypasses that protection entirely. Worth at least a doc note, ideally a same-basename check inside `expand_inputs`/`is_notepack_segment`.

ops/scripts/run-parquet-archive-campaign.sh
```diff
@@ -0,0 +1,285 @@
...
+        fi
+        sync -f "$partial"
+        mv "$partial" "$local_input"
+        sync -f "$INPUT_DIR"
+    fi
+
+    source_sha256="$(sha256sum "$local_input")"
+    source_sha256="${source_sha256%% *}"
+    work_unit_id="notepack-sha256-$source_sha256"
+    "$CAMPAIGN_BIN" "$local_input" \
```
erskingardner: This script runs under `set -euo pipefail` (line 10) and invokes `$CAMPAIGN_BIN` once per file inside the `while read filename` loop (line 194) without `--continue-on-error`. `pensieve_parquet::scan_framed_notepack` intentionally aborts (rather than quarantining) on *structural* framing corruption, by design (`crates/pensieve-parquet/src/segment.rs:99-102`: "Structural framing errors still abort because the next frame boundary cannot be recovered safely"). That `Err` propagates through `run_notepack_work_unit` and makes the campaign binary exit non-zero, which — because of `set -e` — kills this whole script before it reaches any later filename in the loop, not just the corrupt one. With `Restart=on-failure`/`RestartSec=30s` in the systemd unit, a single structurally-corrupt historical segment (plausible across years of archive data, e.g. a torn segment from a past ingest crash) causes an indefinite restart-loop stuck on that same file, since it's always the lowest unprocessed item — no forward progress on any later segment until an operator intervenes. The `--continue-on-error` flag exists in the binary precisely for this class of problem but isn't used here or documented anywhere under ops/. Given the PR explicitly advertises "quarantine of bad work" as a headline property, this whole-file corruption case seems worth handling (e.g. treat structural corruption as its own quarantine class, or have this script catch a failed unit and skip to the next filename).

crates/pensieve-ingest/src/main.rs
```diff
@@ -1223,5 +1312,107 @@ fn init_pipeline(args: &Args) -> Result<PipelineComponents> {
...
+    let parquet_shadow_handle = if let Some(receiver) = parquet_shadow_receiver {
+        let config = parquet_shadow_config(args)?;
+        tracing::info!(
+            state_db = %config.state_db.display(),
+            staging_dir = %config.campaign.staging_dir.display(),
+            replay_existing = config.replay_dir.is_some(),
+            "starting Parquet shadow worker"
+        );
+        Some(
+            start_parquet_shadow(config, receiver)
```
erskingardner: `start_parquet_shadow(config, receiver).with_context(...)?` runs inside `init_pipeline`, which is called (main.rs:396) before the relay manager is set up and before any relay connection. If this fails — e.g. `Inventory::open(&config.state_db)` or `LocalObjectStore::new(root)`'s `fs::create_dir_all` on a bad `--parquet-shadow-state-db`/`--parquet-shadow-lake-dir` path — the `?` propagates all the way out of `main()` and the whole ingester exits before writing a single notepack segment or connecting to a relay. That's a real coupling between the opt-in shadow feature and the archive-of-record path, contradicting the PR's stated goal that shadow issues shouldn't affect the notepack archive. (Note: this is startup-only — I verified post-startup failures, e.g. an unreachable S3 target, are cleanly isolated and don't affect the notepack path.) Consider treating shadow-init failure as a non-fatal warning that disables shadowing for that run instead of aborting the process.

crates/pensieve-lake/src/work_unit.rs
```diff
@@ -0,0 +1,470 @@
...
+                        .ok_or(Error::NumericOutOfRange {
+                            field: "cleanup.bytes",
+                        })?;
+            }
+            Err(error) if error.kind() == std::io::ErrorKind::NotFound => {}
+            Err(error) => return Err(error.into()),
+        }
+    }
+
+    match fs::remove_dir(&work_dir) {
```
erskingardner: This `fs::remove_dir(&work_dir)` only tolerates `NotFound`. If the process was SIGKILL'd/power-lost mid-write earlier, `ensure_parquet_part`/`ensure_reject_part` (lines ~372-399) may have left a `tempfile`-created file/dir behind in `work_dir` — those rely on `Drop` for cleanup, which never runs on a hard kill, and nothing here sweeps orphaned entries before this call. A single stray leftover makes `remove_dir` fail with a non-empty-directory error even though every artifact this function actually tracks was already removed in the preceding loop, permanently blocking cleanup (and, by CLI default with `continue_on_error=false`, aborting the batch) for that work unit. Worth either globbing/removing unrecognized entries in `work_dir` before this call, or tolerating `remove_dir`'s not-empty error here specifically.

crates/pensieve-parquet/src/validator.rs
```diff
@@ -0,0 +1,334 @@
...
+fn validate_column_profiles_and_pages(file_reader: &SerializedFileReader<File>) -> Result<()> {
+    for row_group_index in 0..file_reader.num_row_groups() {
+        let row_group_metadata = file_reader.metadata().row_group(row_group_index);
+        for (column_index, column) in row_group_metadata.columns().iter().enumerate() {
+            if !matches!(column.compression(), Compression::ZSTD(_)) {
+                return Err(Error::InvalidFile(format!(
+                    "row group {row_group_index} column {column_index} is not Zstandard-compressed"
+                )));
+            }
+            for encoding in column.encodings() {
```
erskingardner, line 202: This forbidden-encoding check reads `column.encodings()`, which is the row group's footer-reported summary (`ColumnMetaData.encodings`), not derived from actual page bytes. The loop right below (lines 214-224) already decodes real `Page`s via `get_column_page_reader` and each `Page::DataPage`/`DataPageV2`/`DictionaryPage` variant carries its own genuine `encoding` field parsed from that page's own header — but only `DataPageV2` is checked there (for the V1-only rule) and the per-page `encoding` field itself is never inspected. A file whose footer claims only `PLAIN`/`RLE`/`RLE_DICTIONARY` while actual pages use a different encoding (e.g. `DELTA_BINARY_PACKED`) would pass this "strict" validator despite violating the compatibility-profile rule (spec §5). Doesn't affect row-level integrity (ID/signature are independently re-verified elsewhere), but for a validator whose whole purpose is to police untrusted files without trusting their self-reported metadata, this one check is trusting exactly the kind of self-reported metadata it's meant to catch lying. Suggest reading `encoding` off each `Page` in the existing page loop instead.

crates/pensieve-parquet/src/writer.rs
```diff
@@ -0,0 +1,289 @@
...
+
+fn prepare_events_with_count<'a>(
+    events: impl IntoIterator<Item = &'a Event>,
+) -> Result<(Vec<CanonicalEvent>, usize)> {
+    let mut by_id = BTreeMap::<[u8; 32], CanonicalEvent>::new();
+    let mut input_events = 0;
+
+    for event in events {
+        input_events += 1;
+        let candidate = CanonicalEvent::from_event(event)?;
```
erskingardner: `CanonicalEvent::from_event(event)?` propagates any single event's validation error out of the whole `prepare_events_with_count`/`write_events` call, so one invalid event (or one bad signature variant of an otherwise-duplicate ID) fails the entire batch rather than being discarded as `docs/parquet_archive_format.md` §3.1 specifies ("invalid signature variants are discarded"). `tests/canonical_writer.rs`'s `rejects_invalid_events_before_writing_bytes` only covers the single-invalid-event case, not the mixed valid+invalid-duplicate-signature case the spec text is describing. Not currently reachable from any caller in this PR — `pensieve-lake` goes through the separate `scan_segment`/`prepare_canonical_events` path, which does discard invalid frames correctly — but this is a public function on the crate with a spec-inconsistent contract, worth aligning before something else calls it.

erskingardner (commented):
(Posted as COMMENT, not REQUEST_CHANGES, because GitHub does not allow the PR author to request changes on their own PR via the API — treat items 1-5 below as blocking in spirit.)

Big, well-tested foundation for the lakehouse migration, and the risky parts are explicitly gated (shadow path off by default, campaign service not installed/enabled by this PR, ClickHouse projection drop kept out of `ch-migrate-all`). I split the diff across four focused passes (pensieve-parquet, pensieve-lake, pensieve-ingest shadow integration, ops/migrations/docs) and independently verified every finding below by reading the actual code rather than trusting the sub-review summaries.

The ClickHouse projection removal/rollback (the highest-risk item by nature) is genuinely solid: idempotent, guarded against concurrent mutations, disk-headroom-gated, and the rollback SQL reconstructs the exact prior projection definition. `just ch-migrate-all`'s glob does exclude `docs/migrations/prepared/` as claimed.

Requesting changes for a handful of concrete, verified issues before this merges — none are data-corruption risks in the paths this PR actually activates today, but a couple directly undercut headline claims in the PR description ("quarantine of bad work", isolation of the shadow path from the archive path) and one will break the campaign service the moment someone tries to use it:

1. **Ops path mismatch** (`ops/systemd/pensieve-parquet-campaign.service:14`, `ops/scripts/run-parquet-archive-campaign.sh:80`, `ops/parquet-campaign.env.example:11`): all three new files default to `/srv/pensieve/repo/...`, but the rest of the repo (see `ops/RUNBOOK.md:3`) consistently uses `/home/pensieve/pensieve`. As shipped, the service fails to start out of the box.
2. **Whole-segment corruption can wedge the unattended campaign indefinitely** (`ops/scripts/run-parquet-archive-campaign.sh`, the per-file loop starting at line 194 / binary invocation at line 252): structural notepack framing corruption intentionally aborts the work unit (`crates/pensieve-parquet/src/segment.rs:99-102`), the wrapper script runs under `set -euo pipefail` without `--continue-on-error`, so one bad historical segment kills the whole run and, combined with `Restart=on-failure`, restart-loops forever on the same file with no forward progress on later segments.
3. **`pensieve-parquet-campaign`'s own directory-mode CLI can defeat duplicate-representation filtering** (`crates/pensieve-lake/src/bin/pensieve-parquet-campaign.rs:174`): `expand_inputs` dedups paths with a plain `Vec::dedup()` (consecutive exact-path only) and work-unit identity is `notepack-sha256-<source file bytes>` (`work_unit.rs:104`), so `.notepack` and `.notepack.gz` copies of the same segment hash differently and both get converted/published. The "prefer gzip, skip the paired plain file" logic lives only in the ops wrapper's awk heuristic, not in the crate.
4. **Enabling the Parquet shadow flag can take down the entire ingester at startup**, not just the shadow feature (`crates/pensieve-ingest/src/main.rs:1315-1326`, `init_pipeline`): `start_parquet_shadow(...)?` runs before the relay manager is constructed, so a bad `--parquet-shadow-state-db`/`--parquet-shadow-lake-dir` path (disk full, permissions, typo) makes `main()` return early via `?` and the process never connects to a relay or writes a single notepack segment — contradicting the PR's own stated invariant that shadow issues shouldn't affect the archive path. (I confirmed steady-state failure isolation — post-startup S3/publish failures — works exactly as intended; this is specifically a startup-time coupling gap.)
5. **`crates/pensieve-lake/src/work_unit.rs` (`cleanup_published_local_artifacts`, ~lines 160-236)**: on a real process kill (SIGKILL/power loss) mid-write, orphaned `tempfile` staging artifacts in `work_dir` are never swept before/around cleanup (only `Drop`-based removal, which doesn't run on SIGKILL). `fs::remove_dir(&work_dir)` at the end only tolerates `NotFound`, so a stray leftover file makes this call fail permanently for that work unit — the tracked artifacts are already gone, but disk for that unit can never be reclaimed and (by CLI default) aborts the batch. Undercuts the "bounded-disk" claim under an actual crash, as opposed to a clean shutdown.

Two more, lower severity / non-blocking, worth a look but not asking you to fix before merge:

6. `crates/pensieve-parquet/src/validator.rs`, `validate_column_profiles_and_pages` (~lines 202-224): the forbidden-encoding check reads `column.encodings()` off the row group's footer metadata (the writer's self-reported summary), not the real per-page `encoding` field already available in the very next loop over decoded `Page`s in the same function. A footer that lies about which encodings it used (while pages actually use e.g. `DELTA_BINARY_PACKED`) would pass as a conformant V1 file. Doesn't affect event ID/signature integrity (those are independently re-verified elsewhere), just weakens the "strict validator" claim for the Parquet compatibility-profile rule specifically.
7. `crates/pensieve-parquet/src/writer.rs`, `prepare_events_with_count` (~lines 105-124): one invalid event anywhere in the batch (`CanonicalEvent::from_event(event)?`) fails the whole `prepare_events`/`write_events` call, rather than discarding just that event/signature-variant as `docs/parquet_archive_format.md` §3.1 describes ("invalid signature variants are discarded"). Currently unused by any caller in this PR (`pensieve-lake` uses the separate quarantine-aware `scan_segment`/`prepare_canonical_events` path instead), so it's a latent public-API contract gap rather than a live bug — worth fixing before anything wires up to it.

PR #32, opened 2026-07-27
Make live Parquet shadow bootstrap safe

erskingardner:
## Summary

- require an explicit Parquet shadow replay policy instead of implicitly replaying every sealed segment
- add an inclusive `--parquet-shadow-replay-from-segment` boundary and verify that first activation starts at exactly the next segment number
- persist the replay policy, publisher identity, object prefix, writer limits, and staging path as immutable SQLite inventory settings
- expose shadow configuration through an optional restricted systemd environment file
- document the production activation, high-water mark, rollback, and historical catch-up procedure

## Why

The historical archive conversion is already running separately. Enabling the existing live shadow against an empty inventory would scan roughly 7,700 pre-existing notepack segments and duplicate that campaign. A missed or changed bootstrap boundary could also create a restart gap.

This change makes the historical/live boundary explicit and durable. Installing the code and systemd unit alone remains inert because `/etc/pensieve/parquet-shadow.env` is optional; the normal notepack and ClickHouse paths continue unchanged until an operator deliberately activates the shadow.

## Operational impact

The intended rollout remains two-stage:

1. deploy and restart the ordinary ingester with no shadow environment file;
2. after that soak is healthy, stop ingestion, calculate `max(existing segment) + 1`, install the restricted shadow configuration, and validate the first published live segment before leaving the shadow enabled.

A shadow startup or publication failure cannot fail the authoritative notepack path.

## Validation

- `just precommit`
- focused restart test proves a live segment is replayed after restart while an older historical segment is excluded
- inventory tests prove immutable settings survive reopen and reject configuration drift
- CLI tests prove a publication target requires an explicit replay policy

### DIFF
--- a/crates/pensieve-ingest/src/pipeline/parquet_shadow.rs
+++ b/crates/pensieve-ingest/src/pipeline/parquet_shadow.rs
@@ -17,6 +17,8 @@ use pensieve_lake::{
 
 use super::SealedSegment;
 
+const REPLAY_POLICY_SETTING: &str = "parquet_shadow.replay_policy.v1";
+
 /// Publication target for the live shadow sink.
 #[derive(Clone, Debug)]
 pub enum ParquetShadowPublisher {
@@ -40,6 +42,12 @@ pub struct ParquetShadowConfig {
     pub publisher: ParquetShadowPublisher,
     /// Directory scanned for sealed work units before consuming live notifications.
     pub replay_dir: Option<PathBuf>,
+    /// Inclusive segment-number floor for startup replay.
+    ///
+    /// This lets a new live shadow take ownership of future segments without
+    /// replaying the pre-existing historical archive. The effective replay
+    /// policy is persisted in the inventory and cannot drift across restarts.
+    pub replay_from_segment: Option<u64>,
 }
 
 /// Start the shadow worker and wait until its inventory and publisher initialize.
@@ -91,8 +99,15 @@ struct ShadowWorker {
 
 impl ShadowWorker {
     fn new(config: ParquetShadowConfig) -> Result<Self> {
-        let inventory = Inventory::open(&config.state_db)
+        let mut inventory = Inventory::open(&config.state_db)
             .with_context(|| format!("failed to open {}", config.state_db.display()))?;
+        let replay_policy = replay_policy(&config);
+        let policy_is_recorded = inventory
+            .setting(REPLAY_POLICY_SETTING)
+            .context("failed to load durable Parquet shadow replay policy")?
+            .is_some();
+        validate_replay_floor(&config, policy_is_recorded)?;
+        ensure_inventory_settings(&mut inventory, &config, &replay_policy)?;
         let publisher: Box<dyn Publisher> = match &config.publisher {
             ParquetShadowPublisher::Local { root } => Box::new(
                 LocalObjectStore::new(root)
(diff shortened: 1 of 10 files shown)


ops/parquet-shadow.env.example
```diff
@@ -0,0 +1,34 @@
...
+PENSIEVE_PARQUET_SHADOW_MAX_EVENT_BYTES=16777216
+
+# Five minutes keeps live objects reasonably current without making event time
+# part of the file boundary.
+PENSIEVE_PARQUET_SHADOW_MAX_BATCH_AGE_SECS=300
+
+# REQUIRED for the first production activation. While the ingester is stopped,
+# set this to one greater than the highest pre-existing segment number. The
+# value becomes an immutable setting in the live shadow inventory.
+PENSIEVE_PARQUET_SHADOW_REPLAY_FROM_SEGMENT=replace-with-next-segment-number
```
erskingardner: The example ships an unparsable placeholder for a typed numeric field. `parquet_shadow_replay_from_segment` is `Option<u64>` and is populated via clap's `env = "PENSIEVE_PARQUET_SHADOW_REPLAY_FROM_SEGMENT"`. If this file is installed and the operator forgets to replace this one line (easy to do, since the other placeholders like `S3_BUCKET=replace-me` are valid strings that parse fine), `Args::parse()` fails on the non-numeric value and the whole `pensieve-ingest` process exits before `main()` body runs (verified locally: exit code 2, "invalid digit found in string"). That takes down the authoritative notepack/ClickHouse path too — not just the shadow — which is the exact failure mode this PR says it prevents ("A shadow startup or publication failure cannot fail the authoritative notepack path").\n\nLeaving the line unset entirely does not have this problem: unset -> `None` -> the existing "choose exactly one Parquet shadow replay policy" error is caught inside `parquet_shadow_config` and only disables the shadow (matches the `shadow_target_requires_an_explicit_replay_policy` test). Suggest shipping this line commented out by default (`# PENSIEVE_PARQUET_SHADOW_REPLAY_FROM_SEGMENT=`) rather than a non-numeric placeholder, and letting the RUNBOOK instruct uncommenting it with the real value.

crates/pensieve-ingest/src/pipeline/parquet_shadow.rs
```diff
@@ -241,6 +379,115 @@ mod tests {
...
+            "segment-000000042.notepack",
+            "segment-000000042.notepack.gz",
+            "segment-000000043.notepack.gz",
+            "other-000000100.notepack.gz",
+        ] {
+            fs::write(directory.path().join(name), []).unwrap();
+        }
+
+        let found = discover_sealed_segments(directory.path(), Some(42)).unwrap();
+        let names: Vec<_> = found
```
erskingardner, line 397: Minor, non-blocking: `next_segment_number`/`segment_number` here duplicate the directory-scan-and-parse logic already in `SegmentWriter::find_next_segment_number` (crates/pensieve-ingest/src/pipeline/segment.rs:228-259), including hardcoding the `"segment-"` prefix instead of reusing `SegmentConfig::segment_prefix`. Not a bug today since the prefix is always `"segment"`, but the two copies can silently drift if that ever changes. Could be worth extracting a shared helper, but not blocking this PR.

erskingardner (commented):
Reviewed the diff (main.rs CLI wiring, parquet_shadow.rs replay-floor logic, pensieve-lake inventory settings, docs/ops). The core design is sound: requiring an explicit replay policy, persisting it as an immutable inventory setting, and validating the inclusive segment-number floor on first activation are all well thought out and well tested (restart-replay test, drift-rejection test, inventory reopen test all check out — `cargo test -p pensieve-ingest -p pensieve-lake` passes locally).

One real problem found (see inline comment), otherwise clean — GitHub won't let me use "Request changes" on your own PR, but treat the finding below as blocking:

- **`ops/parquet-shadow.env.example` line 29**: `PENSIEVE_PARQUET_SHADOW_REPLAY_FROM_SEGMENT=replace-with-next-segment-number` is a non-numeric placeholder for a field typed `Option<u64>` and sourced via clap `env`. If an operator installs the file and misses this one line, `Args::parse()` fails and takes down the *entire* `pensieve-ingest` process (verified locally: clap exits 2 on a non-numeric env value) — including the authoritative notepack path — which is exactly the failure mode this PR is meant to prevent. Leaving the var unset instead (`None`) degrades gracefully through the existing "choose exactly one replay policy" check. Suggest shipping the line commented out by default rather than a placeholder that fails to parse.

Also left a non-blocking note about minor duplication between `next_segment_number` in `parquet_shadow.rs` and `SegmentWriter::find_next_segment_number` in `segment.rs`.

PR #36, opened 2026-07-28
Freeze and audit the bounded historical campaign

erskingardner:
## Summary

- freeze the finite historical notepack source universe at H = 7702 in an immutable, content-addressed manifest
- make every catch-up pass consume and verify that manifest instead of re-listing a moving remote
- add deterministic campaign completion accounting against the full inventory and active-raw catalog view
- narrow the historical validation gate to exhaustive coverage/accounting plus targeted seven-field comparisons

## Operational safety

The currently running campaign is unchanged. The new wrapper is intended for the next catch-up pass after the existing sequential pass exits. A Linux inventory-only smoke test on the conversion VM produced 7,694 bounded sources through segment 7702 and left the active service running.

## Validation

- `just precommit`
- `bash -n ops/scripts/run-parquet-archive-campaign.sh`
- real Storage Box `rclone lsjson` manifest build and verification
- completion audit against a consistent SQLite backup of the live campaign journal
- isolated Linux wrapper inventory-only smoke test on the conversion VM

### DIFF
--- a/crates/pensieve-lake/src/source_manifest.rs
+++ b/crates/pensieve-lake/src/source_manifest.rs
@@ -1,0 +1,721 @@
+//! Frozen historical-source manifests and completion accounting.
+
+use std::collections::{BTreeMap, BTreeSet};
+use std::fs::{self, File};
+use std::io::Write;
+use std::path::Path;
+
+use serde::{Deserialize, Serialize};
+use sha2::{Digest, Sha256};
+
+use crate::{Error, Result, WorkState, WorkUnitRecord};
+
+/// Format identifier for the historical notepack source manifest.
+pub const HISTORICAL_SOURCE_MANIFEST_FORMAT: &str = "pensieve.historical-source-manifest.v1";
+
+/// One exact source object selected for the bounded historical campaign.
+#[derive(Clone, Debug, Deserialize, Eq, Ord, PartialEq, PartialOrd, Serialize)]
+pub struct HistoricalSourceEntry {
+    /// Logical segment number parsed from the canonical filename.
+    pub segment_number: u64,
+    /// Exact source object name, without a host-local path.
+    pub source_name: String,
+    /// Exact remote object bytes observed when the manifest was frozen.
+    pub source_bytes: u64,
+}
+
+/// Aggregate source-manifest totals.
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub struct HistoricalSourceTotals {
+    /// Number of selected logical source segments.
+    pub sources: u64,
+    /// Sum of selected source-object bytes.
+    pub source_bytes: u64,
+}
+
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+struct HistoricalSourcePayload {
+    format: String,
+    max_segment_number: u64,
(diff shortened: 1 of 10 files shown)


crates/pensieve-lake/src/source_manifest.rs
```diff
@@ -0,0 +1,721 @@
...
+            Error::InvalidSourceManifest("source byte total overflows u64".to_owned())
+        })
+    })?;
+    Ok(HistoricalSourceTotals {
+        sources,
+        source_bytes,
+    })
+}
+
+fn canonical_json(value: &impl Serialize) -> Result<Vec<u8>> {
```
erskingardner, line 545: Non-blocking: `canonical_json()` and `content_id()` here are byte-for-byte identical to the private helpers of the same name already in `catalog.rs` (`canonical_json` at catalog.rs:389, `content_id` at catalog.rs:726). Worth factoring into a shared `pub(crate)` helper instead of copy-pasting, per the project's push for minimal duplication.

crates/pensieve-lake/src/source_manifest.rs
```diff
@@ -0,0 +1,721 @@
...
+    if totals.active_raw_rows != totals.output_rows {
+        problems.push(CompletionProblem {
+            source_name: "active-raw-catalog".to_owned(),
+            reason: format!(
+                "active raw row total differs from published output accounting: catalog={} \
+                 inventory={}",
+                totals.active_raw_rows, totals.output_rows
+            ),
+        });
+    }
```
erskingardner, line 389: Non-blocking: the missing/failed/unexpected-work problem paths are covered by `audit_reports_missing_failed_and_unexpected_work`, but the byte-size-mismatch, duplicate-work-units-per-source, non-UTF-8-filename, and active-raw-row-mismatch branches above (roughly lines 330-388) aren't directly exercised by a test. Since this function is the data-integrity gate for the whole campaign, a bit more direct branch coverage would be cheap insurance.

erskingardner (commented):
Solid, well-scoped change. It replaces the fragile awk-based source selection with a well-tested Rust module (`HistoricalSourceManifest` + `audit_historical_completion`), follows the existing content-addressed/canonical-JSON pattern already established in `catalog.rs`, and the shell wrapper changes are a faithful, safer re-implementation of the previous selection logic (freezing remote byte sizes instead of re-querying at download time is a real correctness improvement). No blocking issues found (GitHub won't allow an APPROVE event on your own PR, so this is posted as a comment-only review meant as an approval). Two minor, non-blocking notes below as inline comments.

- The gzip-preference/high-water-mark selection logic in `source_manifest.rs` correctly mirrors the old awk script's invariants (a plain file is only eligible below the true remote gzip high-water mark), exercised by targeted unit tests.
- `write_historical_source_manifest_noclobber` / `read_historical_source_manifest` correctly enforce immutability and canonical-byte round-tripping, consistent with `ops/RUNBOOK.md`'s "never delete/replace a populated manifest" guidance.
- The wrapper script (`ops/scripts/run-parquet-archive-campaign.sh`) correctly drops the old per-file `rclone size` call in favor of the frozen manifest size, still validates the actual download against that size, and removes the now-redundant `sort -V` since manifest entries are already emitted in ascending segment order.

PR #37, opened 2026-07-28
Add evidence-preserving archive recovery

erskingardner:
## Summary

- salvage only terminally truncated notepack inputs into an atomic evidence bundle containing every complete frame, the exact incomplete tail, and a content-addressed report
- bind a failed original source to a published salvage repair through a no-clobber exception ledger used by the historical completion audit
- promote exact-ID relay recovery into a resumable first-class binary with operator-supplied relay sets, strict output validation, durable journaling, and atomic missing-ID snapshots
- document the damaged-source and recurring segment-7703 recovery procedures

## Real segment 1151 result

The conversion VM retained and processed the actual 201,229,929-byte source. It contains 160,300 valid complete frames followed by only a four-byte length prefix declaring a missing 31,827-byte payload. The complete prefix was published through an isolated repair inventory as one strictly validated 160,300-row Parquet object. The exception ledger now removes segment 1151 from the live completion-audit problem set while preserving the failed original row. The main campaign remained active.

## Stack

This draft targets `codex/bounded-campaign-manifest` and should merge after #36. After #36 lands, retarget this PR to `master`.

## Validation

- `just precommit`
- focused salvage, exception-accounting, and relay-recovery tests
- real source download/checksum and terminal-tail evidence capture
- real isolated S3 repair publication and strict Parquet validation
- real exception-ledger build and progress audit against the live campaign inventory

### DIFF
--- a/crates/pensieve-ingest/src/bin/recover-events.rs
+++ b/crates/pensieve-ingest/src/bin/recover-events.rs
@@ -1,0 +1,379 @@
+//! Resumably recover exact Nostr event IDs from an operator-supplied relay set.
+
+use std::collections::HashSet;
+use std::fs::{self, File, OpenOptions};
+use std::io::{BufRead, BufReader, BufWriter, Write};
+use std::path::{Path, PathBuf};
+use std::time::Duration;
+
+use anyhow::{Context, Result, bail};
+use clap::Parser;
+use futures_util::{StreamExt, stream};
+use nostr_sdk::prelude::{Client, Event, EventId, Filter, JsonUtil};
+
+#[derive(Debug, Parser)]
+#[command(about = "Resumably recover exact Nostr event IDs from public relays")]
+struct Args {
+    /// One 64-character event ID per line.
+    ids: PathBuf,
+    /// Append-only validated recovered-event JSONL.
+    output_jsonl: PathBuf,
+    /// Atomically replaced snapshot of IDs not present in the output.
+    missing_ids: PathBuf,
+    /// Relay URLs, one per line; blank lines and # comments are ignored.
+    #[arg(long)]
+    relay_file: PathBuf,
+    /// Append-only operator journal for batch outcomes.
+    #[arg(long)]
+    journal: Option<PathBuf>,
+    /// Progressively smaller exact-ID request sizes.
+    #[arg(long, value_delimiter = ',', default_value = "100,20,1")]
+    batch_sizes: Vec<usize>,
+    /// Parallel relay-pool requests.
+    #[arg(long, default_value_t = 4)]
+    concurrency: usize,
+    /// Timeout for one batch request.
+    #[arg(long, default_value_t = 12)]
+    request_timeout_secs: u64,
+    /// Relay connection warm-up before requests begin.
+    #[arg(long, default_value_t = 5)]
+    connect_wait_secs: u64,
+}
+
+fn main() {
+    if let Err(error) = run() {
+        eprintln!("event recovery failed: {error:#}");
+        std::process::exit(1);
+    }
+}
+
+#[tokio::main]
+async fn run() -> Result<()> {
+    let args = Args::parse();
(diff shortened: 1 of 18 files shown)


crates/pensieve-ingest/src/bin/recover-events.rs
```diff
@@ -0,0 +1,385 @@
...
+            let mut added = 0usize;
+            match result {
+                Ok(events) => {
+                    for event in events {
+                        if !target_ids.contains(&event.id) || recovered.contains(&event.id) {
+                            continue;
+                        }
+                        event.verify().with_context(|| {
+                            format!("relay returned invalid event {}", event.id)
+                        })?;
```
erskingardner: A single event with a matching ID but an invalid signature aborts the whole recovery run: this `?` propagates the verify error out of `run()`, and `main()` (line 52) exits via `std::process::exit(1)`, which skips fsync for any already-added-but-unflushed events from this batch and kills every other in-flight batch from `buffer_unordered`.

Given event IDs only pin content (not the signature), a relay can legitimately serve a byte-correct-ID event with a broken/missing signature without any malice, and this tool is explicitly meant to run unattended across many relays over long/recurring rounds (per the PR description and `docs/segment_7703_recovery.md`). One bad response from any single relay currently kills the entire pass.

The rest of this codebase already has a convention for this exact situation — quarantine/reject the bad item and keep going (`convert_segment_quarantining_invalid`, "Invalid events are never stored" in CLAUDE.md) rather than aborting the whole operation. Suggest logging + skipping the invalid event here instead of propagating with `?`.

crates/pensieve-ingest/src/bin/recover-events.rs
```diff
@@ -0,0 +1,385 @@
...
+                Err(error) => {
+                    eprintln!("worker={worker} failed_relay={relay} error={error}");
+                }
+            }
+        }
+        if added == 0 {
+            bail!("worker {worker} could not add any configured relay");
+        }
+        client.connect().await;
+        clients.push(client);
```
erskingardner, line 93: Non-blocking: this creates `--concurrency` independent `Client`s, each dialing every relay in `relays`, so total connections per relay scale with `--concurrency`. `nostr_sdk::Client` is `Clone` and pool-backed, so a single client (or fewer clients) cloned across the concurrent `fetch_events` calls would likely give the same parallelism without multiplying connections to every relay — worth a look, but not blocking this PR.

erskingardner (commented):
Solid, well-scoped tooling for a real incident (segment 1151 truncation, segment 7703 residual gap). The content-addressed/no-clobber patterns for the salvage bundle and exception ledger follow existing project conventions closely, and the exception-ledger accounting in `audit_historical_completion` is fail-closed in all the right places (byte size, checksum, work state, repair coverage, accounting overflow).

One real robustness bug in `recover-events` (inline comment) undermines the tool's stated "resumable" design goal: a single relay returning one event with a valid ID but an invalid signature aborts the *entire* recovery run instead of being skipped, discarding whatever else was in flight for that pass. Given this tool is meant to run unattended across "recurring rounds" against public, less-trusted relay sets (per the PR description and `docs/segment_7703_recovery.md`), and the project's existing convention elsewhere is to quarantine bad data rather than abort (`convert_segment_quarantining_invalid`, "Invalid events are never stored"), this should skip+log the bad event rather than propagate with `?`.

Two smaller, non-blocking notes left inline:
- `recover-events` spins up `--concurrency` independent `Client`s, each dialing every configured relay, multiplying per-relay connections by the concurrency factor.
- The new exception-ledger/repair-accounting logic in `audit_historical_completion` (`crates/pensieve-lake/src/source_manifest.rs`) only has a happy-path test (`terminal_truncation_exception_requires_and_accounts_for_active_repair`). Given how much fail-closed branching was added (checksum mismatches, missing repair fragment, accounting overflow, conflicting repair coverage), a couple of negative-path tests would pay for themselves here.

PR #38, opened 2026-07-28
Add first DuckDB/Postgres analytics slice

erskingardner:
## Summary

- freeze the current 24-route ClickHouse analytics contract and migration order
- add the first exact DuckDB analytics builder over an explicit active-file snapshot
- deduplicate cross-file event IDs and build overview, daily event, daily-kind, and all-time kind products
- add versioned Postgres serving relations with run/input provenance and atomic current-run publication
- add build-only and local-object modes, an operator runbook, and environment template

## Why

Pensieve needs a rebuildable shadow analytics path before ClickHouse can be retired. This slice establishes the catalog, computation, validation, and publication boundaries without changing `pensieve-serve`, Grafana, ingestion, or the canonical Parquet lake.

The time semantics are versioned as provisional `slice-a-v1` behavior for old/new comparison. This PR does not approve API cutover or deploy Postgres.

## Validation

- `just precommit`
- exact local fixtures covering cross-file duplicates, pre-genesis timestamps, future timestamps, `u32::MAX + 1`, rolling-window boundaries, and an empty active snapshot
- disposable Postgres 17 round trip covering schema creation, bulk publication, reads through current views, and idempotent retry

### DIFF
--- a/crates/pensieve-analytics/src/build.rs
+++ b/crates/pensieve-analytics/src/build.rs
@@ -453,6 +453,9 @@ fn scalar_u64(connection: &Connection, sql: &str) -> Result<u64> {
         .ok_or_else(|| Error::Validation(format!("query returned no scalar row: {sql}")))
 }
 
+// DuckDB does not accept bind parameters in `CREATE SECRET` options or in a
+// `read_parquet` file-list literal. Those values come from the operator-selected
+// catalog and CLI configuration, so quote and escape them in one place.
 fn sql_string(value: &str) -> String {
     format!("'{}'", value.replace('\'', "''"))
 }
--- a/crates/pensieve-analytics/src/build.rs
+++ b/crates/pensieve-analytics/src/build.rs
@@ -1,0 +1,458 @@
+//! Exact Slice A computation in a persistent DuckDB work database.
+
+use std::path::Path;
+
+use duckdb::{Connection, OptionalExt, params};
+use pensieve_core::NOSTR_GENESIS_TIMESTAMP;
+use serde::Serialize;
+
+use crate::{Error, ResolvedSnapshot, Result};
+
+/// Version of the SQL semantics materialized by this crate.
+pub const QUERY_VERSION: &str = "slice-a-v1";
+const API_TIMESTAMP_MAX: u64 = u32::MAX as u64;
+const SEVEN_DAYS_SECS: u64 = 7 * 24 * 60 * 60;
+const THIRTY_DAYS_SECS: u64 = 30 * 24 * 60 * 60;
+const HOURS_PER_SEVEN_DAYS: f64 = 168.0;
+
+/// Reproducible inputs to one full analytics build.
+#[derive(Clone, Debug)]
+pub struct BuildConfig {
+    /// Fixed upper event-time boundary used by rolling/non-future metrics.
+    pub as_of_epoch: u64,
+    /// Operator/build identity recorded with the run.
+    pub code_version: String,
+    /// AWS region used by DuckDB's S3 credential-chain secret.
+    pub s3_region: String,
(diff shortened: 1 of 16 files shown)


crates/pensieve-analytics/src/build.rs
```diff
@@ -0,0 +1,458 @@
...
+}
+
+fn scalar_u64(connection: &Connection, sql: &str) -> Result<u64> {
+    connection
+        .query_row(sql, [], |row| row.get(0))
+        .optional()?
+        .ok_or_else(|| Error::Validation(format!("query returned no scalar row: {sql}")))
+}
+
+fn sql_string(value: &str) -> String {
```
erskingardner, line 459: Non-blocking style note: `configure_remote_access` (CREATE SECRET) and `materialize_canonical_events` (`read_parquet([{paths}])`) build SQL via string interpolation through `sql_string`'s manual quote-escaping, while every other query in this file uses `params![]` bind parameters. Today the inputs are operator-controlled (catalog `store_id`, CLI flags), so this isn't exploitable, but it's worth a short comment noting why hand-escaping is safe here (DuckDB's `CREATE SECRET`/`read_parquet` list literals aren't bindable), so a future reader doesn't have to re-derive that.

erskingardner (commented):
## Summary

Well-scoped, well-tested first slice of the DuckDB/Postgres analytics path. It's purely additive (new `pensieve-analytics` crate + Postgres migration + docs), touches nothing in `pensieve-serve`, ingestion, ClickHouse, or preview, and is explicit that it is not accepted for API cutover or production deployment.

(Note: GitHub won't let me submit an `--approve` review on your own PR, so this is posted as a comment review — the content below is an approve-level recommendation.)

Verified in detail:
- **Cross-file dedup rule** (`row_number() OVER (PARTITION BY id ORDER BY sig)`, smallest sig wins) matches the pre-established canonical rule already used in `pensieve-parquet`'s writer (`docs/parquet_archive_format.md`, `crates/pensieve-parquet/src/writer.rs`) — not a new convention, correctly re-applied at the cross-file layer.
- **Time-domain semantics** (`total_events`/`kind_all_time` include future & API-unrepresentable rows; daily products cap at `u32::MAX` but still include future dates; `latest_event`/`events_7d`/`kinds_30d` cap at `as_of`; `earliest_event` clamps to `NOSTR_GENESIS_TIMESTAMP`) are exactly what `docs/analytics_slice_a.md` documents, and the test fixture (duplicate/pre-genesis/overflow/future events) exercises each boundary and gets the expected numbers.
- **Reconciliation validation** (`validate_rollups`/`validate_sum`, plus matching Postgres `CHECK` constraints) gives a real double-checked invariant rather than an aspirational comment.
- **Atomic publication** (advisory xact lock, `runs`/`current_run` pointer, idempotent retry, rejection of resurrecting a stale run) looks correct, and the `#[ignore]`d Postgres round-trip test covers it.
- Spot-checked the "24 ClickHouse-backed routes" claim in `docs/analytics_endpoint_migration.md` against `crates/pensieve-serve/src/routes/mod.rs` on `master` — it's accurate.

One non-blocking note left inline on `crates/pensieve-analytics/src/build.rs`: `configure_remote_access` and `materialize_canonical_events` build SQL via string interpolation (through `sql_string`'s escaping) rather than the `params![]` bind-parameter style used elsewhere in the same file. Inputs are operator-controlled today so this isn't exploitable, just worth a comment explaining why it's safe.

Also noted (not blocking, not commented inline): the Postgres client is hard-coded to `NoTls`, but this is already flagged prominently in `docs/analytics_slice_a.md` and `ops/analytics.env.example` as a "colocated/localhost only" constraint to be fixed before any networked deployment — appropriate for a shadow-only slice.

No correctness issues found. Recommend merge.

PR #40, opened 2026-09-23
Harden API query load and cache refresh behavior

erskingardner:
Packages the existing API fixes: Postgres time-unit and eligible-kind corrections, temporary 30-second request timeout, bounded ClickHouse query concurrency, and cache refresh protection. Includes the recorded production load-protection notes. This PR does not authorize a deployment. Independent self-reviews and CI must pass before merge.


crates/pensieve-serve/src/bounded_clickhouse.rs
```diff
@@ -0,0 +1,140 @@
...
+        .map_err(|_| ApiError::Overloaded)?;
+    let permit = tokio::time::timeout(std::time::Duration::from_secs(2), permits.acquire_owned())
+        .await
+        .map_err(|_| ApiError::Overloaded)?
+        .map_err(|_| ApiError::Overloaded)?;
+    // Do not free a slot on HTTP cancellation while ClickHouse is still working.
+    tokio::spawn(async move {
+        let _permit = permit;
+        let _slot = slot;
+        work.await.map_err(ApiError::from)
```
erskingardner: The spawned task is what holds the permit, and it waits on `work` with no client deadline. `max_execution_time=20` is cooperative. If ClickHouse does not stop the query, this task keeps one of the two permits until the process exits.

Caller disconnect leaves the HTTP request open, so `cancel_http_readonly_queries_on_client_close` does not run on that path. Wrap `work` in a client-side timeout a little above 20s and drop the future on expiry. Dropping it closes the connection, which is what allows the cancel-on-close setting to fire, and it releases the permit. Surface that expiry as `Overloaded` or a database error.

crates/pensieve-serve/src/cache.rs
```diff
@@ -106,49 +107,82 @@ pub async fn get_or_compute_with_ttl<T, F, Fut>(
...
+                tracing::warn!(key = %key, error = %e, "failed to serialize for cache");
+            }
         }
-    }
 
-    Ok(value)
+        Ok(value)
+    });
+    if let Some(value) = stale {
+        return Ok(value);
```
erskingardner: This returns the stale value and drops the `JoinHandle`. The refresh keeps the per-key lock, but a failed `compute()` only finishes the task. Nothing logs that error, and the next caller starts another refresh, so a persistent failure stays invisible while aggregates keep being served stale.

Log the error inside the spawned refresh before propagating it. That covers this detached path and the awaited path below.

crates/pensieve-serve/src/routes/stats.rs
```diff
@@ -2489,47 +2517,54 @@ pub async fn engagement(
     State(state): State<AppState>,
     Query(params): Query<EngagementQuery>,
 ) -> Result<Json<EngagementStats>, ApiError> {
+    // Temporarily fail explicitly instead of launching an unbounded raw tag scan.
+    // Postgres serving is not selected or enabled by this guard.
+    if !state.uses_postgres(AnalyticsFamily::Engagement) {
+        return Err(ApiError::MetricUnavailable);
```
erskingardner: With this guard in place the ClickHouse branch below is unreachable. Postgres off returns here; postgres on takes the inner `fetch_postgres_engagement` return. The tag scan this change is keeping off the server is still in the handler, so removing the guard, or the inner `if`, turns that scan back on.

Delete the unreachable query and call `fetch_postgres_engagement` directly. Restoring a ClickHouse engagement query should be a later change that uses a bounded aggregate.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-23T13:58:17Z` |
| Commit reviewed | `bd78416c9124be617066fe1e1c4465dfd45fb7a8` |
| Model | `Grok 4.7` |

Admission control, per-key refresh single-flight, and the Postgres hour-ordinal corrections match the contracts in this PR. `serving_hourly_counts.hour_epoch` is the hour ordinal written by `serving_publish` (`complete_through_epoch / 3600`), and rewriting `event_period_expression` with `(hour_epoch * 3600)` lines those count keys up with the unix-second sketch leaves. ClickHouse kind activity includes the current calendar window; Postgres stops at the last completed period. Both behaviors are documented.

The detached ClickHouse task needs a hard client-side deadline before merge. Two queries that never return pin both permits for the life of the process. Caller disconnect leaves those HTTP requests open, so `cancel_http_readonly_queries_on_client_close` only helps after something else closes the connection. Also delete the unreachable engagement scan, and log detached cache-refresh failures.

`rustfmt` has passed. Clippy, doctests, and the three test shards were still pending, so executable validation is incomplete.

The `hour_epoch` string replace matches the current `event_period_expression` shapes. A snapshot assertion on those strings would keep a later edit from splitting count keys and distinct keys.

crates/pensieve-serve/src/routes/stats.rs
```diff
@@ -2489,47 +2517,54 @@ pub async fn engagement(
...
-        // Calculate all metrics from events_local consistently.
-        // A reply is a kind=1 event that has at least one e-tag (references another event).
-        let row: EngagementRow = state
-            .clickhouse
-            .query(&format!(
-                "SELECT
+    let cache = state.cache.clone();
+    let result =
+        get_or_compute_with_ttl(&cache, &cache_key, ttl::TIME_SERIES, move || async move {
+            if state.uses_postgres(AnalyticsFamily::Engagement) {
```
erskingardner: The guard at the top of this handler returns `MetricUnavailable` whenever `uses_postgres(Engagement)` is false. So this closure only runs when that flag is true, and everything after this `if` (the ClickHouse query and the ratio math, about 35 lines) can never execute. `EngagementRow` then has no remaining use.

Suggest reducing the closure to `fetch_postgres_engagement(&state, days).await` and deleting the unreachable branch and `EngagementRow`. If the raw-scan query is needed again later, git history has it. As it stands, the code looks like ClickHouse engagement still works.

crates/pensieve-serve/src/cache.rs
```diff
@@ -106,49 +107,82 @@ pub async fn get_or_compute_with_ttl<T, F, Fut>(
...
+            Err(e) => {
+                tracing::warn!(key = %key, error = %e, "failed to serialize for cache");
+            }
         }
-    }
 
-    Ok(value)
+        Ok(value)
+    });
+    if let Some(value) = stale {
```
erskingardner: When `stale` is `Some`, this drops the `JoinHandle`, so any `Err` from `compute()` (for example `Overloaded`, a ClickHouse timeout, or a Postgres error) is silently discarded. The next request starts another refresh, and that failure is discarded too. Nothing shows up in logs until the stale window closes and callers start getting 5xx.

A one-line fix inside the spawned task, e.g. `let value = compute().await.inspect_err(|e| tracing::warn!(key = %key, error = %e, "cache refresh failed"))?;`, would make background refresh failures visible.

crates/pensieve-serve/src/routes/stats.rs
```diff
@@ -807,12 +825,14 @@ async fn fetch_postgres_events(
         "month" => EventDistinctGrain::Month,
         _ => unreachable!("group_by was validated before caching"),
     };
-    let period_expression = event_period_expression(grain);
+    // Serving facts store hour ordinals; sketch leaves store Unix seconds.
+    let period_expression =
+        event_period_expression(grain).replace("hour_epoch", "(hour_epoch * 3600)");
```
erskingardner: Nit: the output is correct for all three grains, but `.replace("hour_epoch", ...)` is a string rewrite over SQL. If `event_period_expression` is ever changed to reference another `hour_epoch_*` identifier, this breaks silently. Consider passing the column expression in explicitly, e.g. `event_period_expression(grain, "hour_epoch * 3600")`, so both callers state their unit.

crates/pensieve-serve/src/routes/kinds.rs
```diff
@@ -263,41 +266,52 @@ pub async fn kind_activity(
...
+            let rows: Vec<KindTimeSeriesRow> = state
+                .clickhouse
+                .query(&format!(
+                    "SELECT
                     {} AS period,
                     count() AS event_count,
                     uniqExact(pubkey) AS unique_pubkeys
                 FROM events_local
-                WHERE kind = ?
+                WHERE kind = ? AND created_at >= {} AND created_at <= now()
```
erskingardner: Non-blocking, for whoever switches kinds to Postgres later. This ClickHouse path returns the last N calendar periods, including the current partial one, and may contain empty gaps. The Postgres path (`fetch_postgres_kind_activity`) excludes the current period and has no lower bound, so it returns the last N non-empty completed periods. `docs/api_postgres_cutover_matrix.md` notes the difference, but the kinds cutover will visibly change results for sparse kinds.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-23T13:59:06Z` |
| Commit reviewed | `bd78416c9124be617066fe1e1c4465dfd45fb7a8` |

The premise holds. This is a stopgap for real production load, and it is already running (`7ac5121`). The design fits the problem: a small semaphore-bounded client, detached cache refreshes so cancelled HTTP callers don't leak ClickHouse work, and a bounded stale-while-revalidate window. The Postgres hour-ordinal corrections look right. I checked every changed `serving_hourly_counts` predicate against the `BIGINT` schema and the `(product_id, kind, hour_epoch)` index, and the `* 3600` / `/ 3600` conversions are consistent.

GitHub won't let me request changes on my own PR, so this review is posted as a comment. Two small fixes should land before merge:

1. **Dead code in `engagement`.** The new early return means the ClickHouse branch inside the cache closure can never run. Delete it, along with `EngagementRow`, so the handler doesn't read as if ClickHouse engagement still works.
2. **Refresh failures are lost when a stale value is served.** If the detached refresh fails, the error goes nowhere and nothing is logged. When the stale window ends, callers get errors with no earlier sign of trouble. Add a `warn!` inside the spawned task.

Non-blocking:
- `event_period_expression(grain).replace("hour_epoch", ...)` rewrites SQL with string replacement. Pass the column expression in as a parameter instead.
- Kind activity now means different things on each backend. ClickHouse includes the current partial period and returns the last N calendar periods. Postgres excludes the current period and returns the last N non-empty completed periods. The docs note this, but whoever switches kinds to Postgres needs to know clients will see a change.
- The `tokio::spawn` in `admitted()` probably isn't needed today, because every ClickHouse call already runs inside a detached cache refresh. Keeping it for defense in depth is reasonable, but note that choice in a comment.
- The 503 `analytics_busy` response could send a `Retry-After` header.

Validation gap: at review time, `rustfmt` passes. `clippy`, `doctests` and the three test shards were still pending after about 10 minutes, apparently queued. I ran nothing locally.

erskingardner:
One-pass review fixes are in 08e9c06: detached ClickHouse execution now has a hard 25-second client deadline with a stalled-future/drop/permit-release regression; background cache refresh failures are logged; the unreachable raw ClickHouse engagement scan and row type are removed. Full just precommit passed on this branch. No second reviewer round was launched, and these changes are not being described as independently approved. No merge or deployment performed.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-23T14:15:45Z` |
| Commit reviewed | `08e9c060a78818d9c7103596f8c7c160f25ffc67` |

Re-review at `08e9c06`. All three items the earlier reviews flagged as blocking are fixed:
- Detached ClickHouse work now has a hard 25s client deadline. `stalled_work_is_dropped_and_all_slots_are_released` covers the future being dropped and both semaphores being released.
- `cache refresh failed` is now logged inside the detached refresh.
- The unreachable engagement scan and `EngagementRow` are deleted.

Other things I checked that look right:
- Every changed `serving_hourly_counts` predicate converts seconds to hour ordinals, and the stored column stays sargable.
- Parameter types in the Postgres kind-activity CTE are consistent: `$3` is text in both `date_trunc` and `grain`, and `$6::bigint`.
- The ClickHouse week window (`toMonday(now())`) matches the `toMonday(created_at)` grouping.
- Switching from a global `time_to_live(300s)` to per-entry `expires_at` with `time_to_idle` also fixes an existing bug. The global TTL silently cut `STABLE` and `TIME_SERIES` entries short at 5 minutes.

**Serious concern (inline, `cache.rs:130`):** `/stats/overview` is the endpoint most exposed to this admission design, and the stale allowance excludes it. On ClickHouse it fans out four admitted queries against 2 permits with a 2s wait, and its 60s TTL is below the `AGGREGATES` threshold. While any two slow queries hold the permits, overview returns 503 on every refresh. Changing the check to `ttl > ttl::REALTIME` would fix it.

**Non-blocking (inline):**
- A client-deadline hit is reported as `analytics_busy`, which operators can't tell apart from queue pressure.
- `docs/api_clickhouse_load_protection.md` doesn't mention the new deadline. It also doesn't say that the recorded deployment (`7ac5121`) predates this head.

The `hour_epoch` string-replace and `Retry-After` notes from the earlier reviews still apply; I'm not repeating them.

Validation: clippy, rustfmt and doctests pass on this head. All three test shards were still pending at review time. I ran nothing locally.

crates/pensieve-serve/src/cache.rs
```diff
@@ -106,49 +107,84 @@ pub async fn get_or_compute_with_ttl<T, F, Fut>(
...
             .or_insert_with(|| Arc::new(Mutex::new(())))
             .clone()
     };
 
-    let _guard = key_lock.lock().await;
+    // Serve at most one additional TTL of stale aggregate data while a refresh runs.
+    // Never serve stale real-time watermarks.
+    let stale: Option<T> = match cache.get(key).await {
+        Some(entry)
+            if ttl >= ttl::AGGREGATES
```
erskingardner: `ttl::OVERVIEW` is 60s, so `/stats/overview` never gets a stale value, and it is the route that most needs one. On ClickHouse, `fetch_overview` (`stats.rs:186`) uses `tokio::join!` over four admitted queries: total_events, total_pubkeys, total_kinds and earliest_event. A pool of 2 permits with a 2s acquire timeout can't run them together.

Example: two cold `kind_activity` `uniqExact` scans hold both permits, each for up to 25s. The four overview sub-queries wait 2s, return `Overloaded`, and the whole response is a 503. Each request queued on the key lock then gets the lock, still finds no cache entry, and runs the same failing refresh again. So the headline endpoint is down for as long as two slow queries run, which is the load pattern this PR targets. (`active_users_summary` does the same fan-out with three queries, but its 300s TTL does qualify for stale serving.)

Smallest fix:
```suggestion
            if ttl > ttl::REALTIME
```
`/stats/events/latest` still never serves stale data. Overview's `latest_event` would be at most about 2 minutes old instead of 1. If that is unacceptable, cache the ClickHouse overview components separately, the way `earliest_event_scan` already is.

crates/pensieve-serve/src/bounded_clickhouse.rs
```diff
@@ -0,0 +1,180 @@
...
+        .map_err(|_| ApiError::Overloaded)?;
+    // Do not free a slot on HTTP cancellation while ClickHouse is still working.
+    tokio::spawn(async move {
+        let _permit = permit;
+        let _slot = slot;
+        // The server's 20-second deadline is cooperative. Bound connection and
+        // body-read stalls too, even after the original HTTP caller disconnects.
+        tokio::time::timeout(deadline, work)
+            .await
+            .map_err(|_| ApiError::Overloaded)?
```
erskingardner: Non-blocking: a query that hits the 25s client deadline returns `Overloaded`. That is the same `503 analytics_busy` a caller gets when admission is rejected. For clients, retry-later is fine. For operators, a stalled connection or runaway body read now looks exactly like queue pressure. The only log line is the refresh `warn!`, and it prints `analytics capacity exhausted; retry later`. Log a distinct `warn!` here (for example `"ClickHouse client deadline exceeded"`) so deadline hits show up separately from saturation.

docs/api_clickhouse_load_protection.md
```diff
@@ -0,0 +1,54 @@
+# ClickHouse API load protection
+
+This change affects the API client only, not ingestion or ClickHouse server defaults.
+
+- Two executing database calls; sixteen additional callers may wait up to two seconds.
+  Admission failures return HTTP 503 `analytics_busy`.
+- Each query sends `max_threads=4`, `max_execution_time=20`,
+  `timeout_overflow_mode=throw`, and disconnect cancellation. Server execution
+  deadlines are cooperative, not hard wall-clock guarantees.
+- A detached database task retains its permit until completion even if its HTTP
```
erskingardner: This bullet is stale as of 08e9c06. The task now holds its permit until completion *or the 25s client deadline*, and the deadline is what bounds permits against stalled connections. Please add it to this list. Also note that the deployment record below is for `7ac5121`, which predates the deadline, the refresh-failure logging and the dead-code removal. Otherwise readers will assume this doc describes what production runs.

crates/pensieve-serve/src/bounded_clickhouse.rs
```diff
@@ -0,0 +1,180 @@
...
+    let slot = admission
+        .try_acquire_owned()
+        .map_err(|_| ApiError::Overloaded)?;
+    let permit = tokio::time::timeout(std::time::Duration::from_secs(2), permits.acquire_owned())
+        .await
+        .map_err(|_| ApiError::Overloaded)?
+        .map_err(|_| ApiError::Overloaded)?;
+    // Do not free a slot on HTTP cancellation while ClickHouse is still working.
+    tokio::spawn(async move {
+        let _permit = permit;
```
erskingardner: `fetch_overview` starts four ClickHouse reads together (`fetch_total_events`, `fetch_total_pubkeys`, `fetch_total_kinds`, `fetch_earliest_event`), and a fifth (`fetch_latest_event`) when no watermark path is set. `active_users_summary` starts three. Each `fetch_*` acquires its own permit here, while `tokio::join!` launches them at once.

Two of those calls run. The others wait two seconds and return `Overloaded`. The join still waits for the calls that got permits. Those calls sit in the detached task below, so the sibling timeout does not stop them, and the handler drops their rows. Both permits stay taken until the 25-second deadline.

Overview uses a 60-second TTL, under the stale-while-revalidate cutoff, so a failed refresh is not cached. The next request fans out again. For as long as two of those queries take more than two seconds, overview stays on 503 and holds the pool.

Sequence the joins so one request keeps one running query, or make this wait use the same deadline as the query. Keep the two-second bound for separate requests.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-23T14:20:46Z` |
| Commit reviewed | `08e9c060a78818d9c7103596f8c7c160f25ffc67` |
| Model | `Grok 4.7` |

The hour-ordinal serving corrections, the detached cache refresh, and the 25-second ClickHouse deadline are the right stopgap. The unreachable engagement scan and the silent refresh failure from the earlier review are fixed on this commit.

One blocker remains. `fetch_overview` issues four ClickHouse reads at once (five when no watermark path is set), and `active_users_summary` issues three. Each read takes its own permit, and only two permits exist. A sibling that waits longer than two seconds fails the whole `tokio::join!`, the handler discards the queries that did get permits, and those queries keep both permits until the 25-second deadline. Overview's 60-second TTL is below the stale-while-revalidate cutoff, so the failed refresh is not stored and the next request repeats the fan-out. While any two of those queries run longer than two seconds, overview stays on 503 and occupies the pool this PR is adding.

Sequence those joins, or wait for a permit until the query deadline. The two-second bound can stay for separate requests.

Non-blocking: `docs/api_clickhouse_load_protection.md` still says a detached call holds its permit until completion. This commit also drops the call at 25 seconds.

Validation: `rustfmt`, `clippy`, and doctests passed on this commit. The three test shards were still pending.

GitHub will not accept a request-changes review from the PR author, so this is posted as a comment. Treat the blocker above as required before merge.

erskingardner:
Round 2 address pass: 3cd332e sequences overview and active-user summary component reads so a refresh cannot reject its own sibling queries against the two-slot pool. Overview now has at most one additional TTL of stale allowance; real-time values remain excluded. Added a regression for overview stale/expired and real-time behavior, and clarified the historical deployment does not cover later fixes. Full just precommit passed. Revised code has not been independently reviewed again; no merge or deployment. The separate-request two-second queue limit and 25-second client deadline remain intact.

crates/pensieve-serve/src/cache.rs
```diff
@@ -106,49 +107,84 @@ pub async fn get_or_compute_with_ttl<T, F, Fut>(
...
         let mut map = IN_FLIGHT.lock().await;
+        map.retain(|_, lock| Arc::strong_count(lock) > 1);
         map.entry(key.to_string())
             .or_insert_with(|| Arc::new(Mutex::new(())))
             .clone()
     };
 
-    let _guard = key_lock.lock().await;
+    // Serve at most one additional TTL of stale aggregate data while a refresh runs.
+    // Never serve stale real-time watermarks.
```
erskingardner: Nit: this comment no longer matches the predicate. Overview is now stale-eligible, and it embeds the `latest_event` watermark. Consider: "Serve at most one additional TTL of stale data for entries with TTL above `REALTIME` while a refresh runs. `REALTIME` entries never serve stale; overview's embedded watermark may be up to 2× `OVERVIEW` old."

crates/pensieve-serve/src/cache.rs
```diff
@@ -209,6 +244,111 @@ mod tests {
...
+                    tokio::time::timeout(Duration::from_secs(1), request)
+                        .await
+                        .unwrap()
+                        .unwrap(),
+                    1
+                );
+                release.send(()).unwrap();
+            } else {
+                tokio::task::yield_now().await;
+                assert!(!request.is_finished());
```
erskingardner: Nit: this assertion can never fail. On the current-thread test runtime, one `yield_now` doesn't guarantee the spawned request has even started, so `!is_finished()` holds whatever the logic does. The real check is `request.await == 2` below, because a stale return would yield `1`. Drop these two lines. You could also split the table into two small tests (stale returns 1, not-stale returns 2) so `expected` isn't also acting as a control-flow flag.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-23T14:32:13Z` |
| Commit reviewed | `3cd332ec76e9a204444b6e9bbf01dc6ef8874746` |

This re-review covers `3cd332e`. Both round-2 concerns are fixed, and I found nothing blocking.

- **Overview can serve stale data again.** The stale check is now `ttl > ttl::REALTIME`, so `/stats/overview` (60s TTL) can return one extra TTL of stale data while a refresh runs. `latest_event` (`REALTIME`) still never serves stale. The new table test covers the stale case, the expired case and the real-time case.
- **Sibling reads no longer starve each other.** `fetch_overview` and `active_users_summary` now run their reads one after another, so a single refresh holds at most one permit. That fixes the 503 loop where sibling reads blocked each other, and it's simpler than the other option (waiting on a permit until the deadline). I checked `pensieve-serve` for any remaining `join!`, `try_join` or `join_all` fan-outs over ClickHouse reads and found none. The cost is higher cold-overview latency, since it becomes the sum of the reads instead of the slowest one. The doc covers that, and the detached refresh still fills the cache after an HTTP timeout.
- **Docs.** They now describe the 25s client deadline, and they say the `7ac5121` deployment record does not cover later fixes.

Two non-blocking notes are inline: a vacuous assertion in the new test, and a cache comment that no longer matches which entries can be stale. One more small point: the overview's embedded `latest_event` can now be up to `2 × OVERVIEW` (120s) old. That is well inside the 900s default `PENSIEVE_LATEST_EVENT_WATERMARK_MAX_AGE_SECS`, but an operator who sets that below ~120s would not get the fail-closed behavior on `/stats/overview`.

The earlier non-blocking notes still apply and I'm not repeating them: the `hour_epoch` string replace, `Retry-After`, and the deadline error sharing `analytics_busy`.

Validation: `rustfmt` passes on this head. `clippy`, doctests and all three test shards were still running at review time. I ran nothing locally. Merge only after CI is green.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-23T14:39:11Z` |
| Commit reviewed | `3cd332ec76e9a204444b6e9bbf01dc6ef8874746` |
| Model | `Grok 4.7` |

The stopgap is still the right change, and the blockers from the reviews on `bd78416` and `08e9c06` are fixed on this commit. Detached ClickHouse work now drops at 25 seconds and releases both semaphores. Refresh failures are logged. The unreachable engagement scan is gone. Overview and active-user summary no longer fan out against the two-permit pool, and there is no remaining `join!` of admitted queries. Overview can be served for one extra 60-second TTL; `ttl::REALTIME` still cannot. The `hour_epoch` ordinal conversions match `serving_publish` (`complete_through_epoch / 3600`) and the second-based sketch leaves.

`rustfmt` passed on this commit. Clippy, doctests, and the three test shards were still running when this review was posted. I did not re-run them locally.

The earlier non-blocking notes still apply and do not need another round: the `hour_epoch` string replace, and client-deadline failures sharing `analytics_busy` with admission rejection.

GitHub rejected `--approve` because this account opened the pull request. This comment is the merge recommendation.

PR #41, opened 2026-09-23
Harden archive admission and fail closed until recovery

erskingardner:
Archive admission hardening against master (PR40 is merged). Adds independent event verification, owned dedupe reservations, fail-closed archive write/seal recovery markers, resident recovery-required state, and an archive-admission health endpoint for Uptime Kuma. Includes Nostr security patch dependency updates; not a comprehensive dependency audit.

Approved operator policy: recovery is manual and incident-specific. A universal repair tool/procedure is intentionally out of scope and is not a merge prerequisite. Required behavior is safe stopping, preserved evidence, observable failure, and no automatic resumption. Notification delivery through the existing Uptime Kuma instance must be verified at deployment. No production deployment or monitor configuration is part of this PR.

Full just precommit passed on 2178719. Independent review and remote CI remain the merge gates.


crates/pensieve-ingest/src/pipeline/segment.rs
```diff
@@ -222,6 +226,28 @@ impl SegmentWriter {
...
+            .output_dir
+            .join(format!("{}.recovery-required", config.segment_prefix));
+        if marker.exists() {
+            return Err(Error::Segment(format!(
+                "archive recovery required: {}",
+                marker.display()
+            )));
+        }
+        for entry in fs::read_dir(&config.output_dir)? {
+            let path = entry?.path();
```
erskingardner: **Blocking.** This makes every unclean exit an outage that needs an operator. While ingesting there is almost always a current `.open` segment, so an OOM kill, a SIGKILL after `TimeoutStopSec` during shutdown, a host reboot, or a panic in any other task leaves one behind. The unit uses `Restart=always`/`RestartSec=10`, so this becomes a crash loop, which is exactly what the runbook says not to do. Any orphan `.open` files already in production will also block the first start after deploy.

The check doesn't buy the safety it's meant to. The bug is appending after a torn frame and then sealing. After a restart the writer gets a new number from `find_next_segment_number`, which already skips `.open`, so the torn file is never appended to or sealed. Its IDs were never marked `Archived` in dedupe or sync-state, so the events stay re-fetchable. That is how crash orphans worked before this PR. The in-process latch plus the marker already cover a real I/O failure. The disk-full-without-marker case leaves the same state as a crash mid-write, which the old design already handled.

The recovery doc also has nothing to run for this case. "Canonical recovery path" isn't defined, and `pensieve-notepack-salvage` rejects complete inputs, which a clean crash orphan usually is.

Suggestion: gate startup on the marker only and leave bare `.open` orphans as they were, or ship an automated recovery in the same change. `seal_checkpointed` has the same problem: a SIGKILL during the final shutdown seal leaves the marker behind.

crates/pensieve-ingest/src/main.rs
```diff
@@ -1110,26 +1113,23 @@ async fn main() -> Result<()> {
                         error = %compact_error(&e),
                         "dedupe check failed for live event"
                     );
-                    // Treat as novel on a (rare, transient) dedupe error: writing a
-                    // possible duplicate is safe (ClickHouse de-dups by id), whereas
-                    // dropping it would silently lose a potentially-new event.
-                    true
+                    return Err(e);
```
erskingardner: **Blocking.** `RelaySource::run_async` handles a handler `Err` with `break` (`relay.rs` around L618-628), so one transient RocksDB read error now ends live ingestion and runs the full shutdown. The event is still dropped, since a live subscription won't redeliver it, and all other relays go down with it. The negentropy path is fine because `run_periodic` just skips recording that event. Here, fail closed for this event only: log, increment a counter, `return Ok(true)`. No duplicate is written, and negentropy can pick the event up later.

crates/pensieve-ingest/src/pipeline/segment.rs
```diff
@@ -408,6 +461,90 @@ impl SegmentWriter {
...
+        let mut marker = File::create(self.recovery_marker())?;
+        marker.write_all(
+            b"Archive operation interrupted. Inspect and recover before removing this marker.\n",
+        )?;
+        marker.sync_all()?;
+        File::open(&self.config.output_dir)?.sync_all()?;
+        Ok(())
+    }
+
+    fn archive_operation<T>(&self, operation: impl FnOnce() -> Result<T>) -> Result<T> {
```
erskingardner: This lock is held across the whole `seal_checkpointed()`: flush, `sync_all` of the segment on the `/archive` HDD, rename, directory fsync, `mark_archived` of every ID, and the watermark write. `take_current_for_seal` exists so writers can keep going into the next segment during a seal. With this gate, every live and negentropy write now blocks a tokio worker for the whole seal, including the timer seals from `start_parquet_batch_timer`. `current` already serializes frame writes. What's needed is a latch, e.g. a `OnceLock<String>` or an `AtomicBool` plus a reason, that is checked on entry and set on error or panic. Holding a mutex around the operation isn't needed.

crates/pensieve-ingest/src/pipeline/segment.rs
```diff
@@ -336,6 +365,24 @@ impl SegmentWriter {
     ///
     /// Returns `true` if a segment was sealed as a result.
     pub fn write(&self, event: PackedEvent) -> Result<bool> {
+        self.archive_operation(|| self.write_inner(event, None))
+    }
+
+    /// Write a reserved event, releasing its claim if opening a segment fails.
+    /// Once frame I/O begins, failures are ambiguous and retain writer ownership.
+    pub fn write_reserved(
```
erskingardner: Passing the guard into the writer only changes the `ensure_current_segment` failure case. In that case `archive_operation` latches the writer anyway, so the released claim can never be retried in this process. `segment_open_failure_releases_claim_for_retry` shows the release, but no retry is possible. Calling `admission.retain()` right before `segment_writer.write(packed)` at the two call sites gives the same behavior. `write_reserved`, the `Option<PendingAdmission>` parameter and the `super::` coupling can then go.

crates/pensieve-ingest/src/pipeline/admission.rs
```diff
@@ -0,0 +1,106 @@
+//! Validate untrusted SDK events before dedupe or other archive side effects.
+
+use nostr_sdk::Event;
+
+use crate::{Error, Result};
+
+/// Recompute the ID and verify the signature independently of SDK caches.
+pub fn validate_archive_event(event: &Event) -> Result<()> {
```
erskingardner: The module is a single-use wrapper around `event.verify()`, and the ~90 lines of tests under it exercise `DedupeIndex::reserve` and `SegmentWriter`, not validation. Call `event.verify()` directly in `relay.rs` and `negentropy.rs` (`.map_err(DatabaseError::backend)`), and move the tests to `dedupe.rs` and `segment.rs` next to the code they cover.

docs/archive_failure_recovery.md
```diff
@@ -0,0 +1,57 @@
...
+
+1. Inspect logs, disk/inode space, filesystem health and failed paths. Do not
+   delete evidence, clear the dedupe database, or repeatedly restart ingestion.
+2. Stop the ingester before inspecting or recovering mutable archive files.
+   Record the marker, `.open` and any corresponding sealed file identities.
+3. Resolve the underlying storage fault. Preserve an evidence copy before any
+   repair. Enumerate complete length-prefixed frames; validate event IDs and
+   signatures and identify any incomplete trailing frame. Never rename a raw
+   `.open` file to a sealed name merely to make startup pass.
+4. Recover complete valid events through the canonical recovery path, reconcile
```
erskingardner: Nothing in the repo implements "the canonical recovery path". `docs/archive_recovery.md` and `pensieve-notepack-salvage` already exist for truncated sources, but salvage rejects complete inputs, and neither handles a crash-orphan `.open` or a post-rename failure. Please extend `docs/archive_recovery.md` instead of starting a parallel doc, give concrete commands for each marker/`.open` case, and link it from `ops/RUNBOOK.md`, which CLAUDE.md names as the authoritative ops procedure.

docs/negentropy_hardening_plan.md
```diff
@@ -0,0 +1,95 @@
...
+  forged signature, closed receiver, duplicate reservation, unwind, ownership
+  transfer, and segment-open failure.
+
+The updated lockfile audit has 21 remaining vulnerability matches outside Nostr,
+down from 32. This is not a clean whole-workspace security audit. TLS/AWS transport,
+bytes/lz4/time and other findings remain tracked in the original report.
+
+## Verification environment
+
+Use `CARGO_TARGET_DIR=/Volumes/Worktrees/pensieve-negentropy-target` and
```
erskingardner: This file and `negentropy_audit_20260923.md` read as session notes: local build paths, production disk and counter snapshots from one day, SHA-256s of files, and progress checkboxes for later slices on another branch. They will go stale quickly and aren't in scope for an archive-admission PR. Consider moving them to the PR/issue description, and keep only lasting policy in the repo.

ops/production/prometheus/prometheus.yml
```diff
@@ -10,6 +10,9 @@ global:
   scrape_interval: 15s
   evaluation_interval: 15s
 
+rule_files:
```
erskingardner: The rules load, but no `alerting:`/Alertmanager target is configured, so they only fire in the Prometheus UI. This design stops ingestion and waits for a person to act, so an alert nobody receives means a silent outage. Please wire up a receiver, or make it an explicit deploy prerequisite in the RUNBOOK.

Cargo.lock
```diff
@@ -2501,7 +2501,7 @@ dependencies = [
  "libc",
  "percent-encoding",
  "pin-project-lite",
- "socket2 0.6.1",
+ "socket2 0.5.10",
```
erskingardner: Unrelated drive-by: `hyper-util`, `quinn` and `quinn-udp` move from socket2 0.6.1 back to 0.5.10. Please regenerate with only `cargo update -p nostr -p nostr-relay-pool` so the lock diff is just the security bump. Also, `Cargo.toml` pins a floor for `nostr` but not for `nostr-relay-pool`, which is where RUSTSEC-2026-0224/0232 are.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-23T13:51:35Z` |
| Commit reviewed | `fd8e101c32ebdfce2c189d5860fb5e590e61904c` |

Most of this is worth landing: the patched Nostr floor, checking events before dedupe/discovery, the owned pending guard, and latching the writer in-process after a partial frame write. The latch fixes the actual bug the plan describes, which is appending after a torn frame and then sealing a malformed file. Two changes go past that goal and add new outages, so please fix them before merge.

**Blocking**
1. **Startup refuses every `.notepack.open` file** (inline on `segment.rs`). There is almost always an open segment, so an OOM kill, a SIGKILL at the stop timeout, a reboot, or a panic outside `archive_operation` will now stop the ingester from starting. With `Restart=always`/`RestartSec=10` that becomes a crash loop. The recovery doc points to a "canonical recovery path" that doesn't exist for `.open` files. An orphaned `.open` was already harmless: it is never appended to or sealed after a restart, and its IDs are never marked `Archived` in dedupe or sync-state.
2. **A dedupe error now ends live ingestion** (inline on `main.rs`). `run_async` handles a handler `Err` by `break`, so one RocksDB read error shuts down the whole ingester. The event is still lost, because a live subscription doesn't redeliver. The "retry" comment only holds for the negentropy path.

**Non-blocking**
- `admission_gate` holds a mutex for the whole seal (fsync on HDD, `mark_archived`, watermark). This undoes the concurrency that `take_current_for_seal` was built for: live writes now stall during timer and negentropy seals. An atomic latch is enough.
- Passing `PendingAdmission` into the writer does nothing extra. Every `write_inner` error latches the writer, so the claim released on an open failure can never be retried in this process. Retaining at the call site and removing `write_reserved` gives the same behavior.
- `admission.rs` is a 3-line wrapper around `event.verify()`, and its tests exercise `DedupeIndex`/`SegmentWriter`. Inline the call and move the tests.
- About 275 lines of docs are session notes: local `/Volumes/...` paths, production snapshots, and checkboxes for later slices. The recovery policy should extend `docs/archive_recovery.md` and be linked from `ops/RUNBOOK.md`.
- `rule_files` is added without any `alerting:` target. The design depends on a person being notified, and these alerts reach no one.
- Unrelated `socket2` 0.6.1 to 0.5.10 downgrade in `Cargo.lock`. Also, `nostr` gets a floor in `Cargo.toml` but `nostr-relay-pool` doesn't, and the verification-cache advisory is in the relay pool.

CI (clippy/tests/doctests) was still running when this review was written. This review ran nothing locally.

crates/pensieve-ingest/src/pipeline/segment.rs
```diff
@@ -222,6 +226,28 @@ impl SegmentWriter {
         // Create output directory if it doesn't exist
         fs::create_dir_all(&config.output_dir)?;
 
+        let marker = config
+            .output_dir
+            .join(format!("{}.recovery-required", config.segment_prefix));
+        if marker.exists() {
```
erskingardner: A recovery marker or leftover `.notepack.open` returns from `SegmentWriter::new` only after `DedupeIndex::open` in `init_pipeline`. `ops/systemd/pensieve-ingest.service` sets `Restart=always` and `RestartSec=10`, so systemd opens RocksDB again every 10 seconds for a condition that cannot clear itself. `archive_recovery_required` is set only inside `archive_operation`, so this startup path never exports it. The metrics server is already listening while RocksDB opens, so `up{job="ingest"}==0` is not a stable signal for this failure either.

The same loop starts from a latched in-process fault: shutdown calls `segment_writer.seal()?`, that latched error exits the process, and systemd restarts into this check.

Keep the process resident when recovery is required: export `archive_recovery_required=1`, serve `/metrics`, and wait for SIGTERM. Skip another RocksDB open on that path. Pair that with a systemd stop (`RestartPreventExitStatus`, or an equivalent) so a failed init cannot tight-loop. This is the opposite of the runbook line that says not to restart ingestion repeatedly.

crates/pensieve-ingest/src/main.rs
```diff
@@ -1110,26 +1113,23 @@ async fn main() -> Result<()> {
...
             coverage.observe(event);
 
-            if is_novel {
+            if let Some(admission) = admission {
                 // New event - NOW pack it (only for novel events)
                 match pack_nostr_event(event) {
                     Ok(packed) => {
                         // Write to segment
-                        if let Err(e) = segment_writer.write(packed) {
+                        if let Err(e) = segment_writer.write_reserved(packed, admission) {
```
erskingardner: After the writer latches, `write_reserved` returns before `write_inner` and drops this `PendingAdmission`, which removes the id from the in-memory set. The handler logs the error and still returns `Ok(true)`. The next delivery reserves the same id again, and `record_event` above counts `admission.is_some()` as novel. For the whole outage, relay novel-rate climbs toward 100% and every subsequent event emits an error log.

Stop live intake once archive admission is blocked, and keep the process up. Returning `Err` from this arm breaks the relay loop and falls through to `seal()?`, which is what starts the restart loop described on `SegmentWriter::new`.

crates/pensieve-ingest/src/pipeline/dedupe.rs
```diff
@@ -74,7 +74,43 @@ pub struct DedupeIndex {
     pending: Mutex<HashSet<[u8; 32]>>,
 }
 
+/// An exclusive in-memory admission claim, released on pre-admission failure.
```
erskingardner: Non-blocking: the crash-safety comment on `DedupeIndex` still says an unsealed segment's bytes are lost on crash and those events are re-fetched on the next start, because the pending set is memory-only. Startup now refuses while a `.notepack.open` file or recovery marker exists, and dropping the writer flushes the `BufWriter` into that open file. Update that comment so the next reader follows `docs/archive_failure_recovery.md` (preserve the open file, reconcile, then clear the marker) instead of deleting the segment to force a re-fetch.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-23T13:57:28Z` |
| Commit reviewed | `fd8e101c32ebdfce2c189d5860fb5e590e61904c` |
| Model | `Grok 4.7` |

The admission guard, independent `Event::verify()`, and the `nostr-relay-pool` 0.44.3 bump match this integrity slice. In nostr 0.44.7, `Event::verify` recomputes the id and checks the signature, so a forged body that reuses a cached id fails at this boundary.

The fail-closed latch does not survive the service lifecycle. A marker or leftover `.notepack.open` makes the process exit during init, and `Restart=always` opens RocksDB again every 10 seconds. That path never sets `archive_recovery_required`. After the latch, the live handler also keeps running and releases later reservations, so relay novelty stats treat the outage as fresh traffic.

`rustfmt` passed. Clippy, doctests, and the three test shards were still in progress at review time; I did not run tests locally.

GitHub rejected `REQUEST_CHANGES` because this review is from the PR author, so this is filed as a comment. Treat the findings below as blocking.

erskingardner:
One-pass disposition: retaining the explicitly approved fail-closed policy; not replacing storage errors with silent continuation or ignoring orphan archives. b8b560b corrects the crash-safety documentation, removes the misleading implied canonical recovery procedure, and links explicit deployment prerequisites from RUNBOOK. Full just precommit passed. This PR remains draft and unmerged: recovery-required startup/steady-state observability, stopping live intake without a restart loop, tested orphan/post-rename recovery procedures, and an operator-selected/tested notification receiver remain blockers. Documentation is not claimed to fix those runtime issues.

ops/RUNBOOK.md
```diff
@@ -29,6 +29,21 @@ provisioning). The ingester exposes Prometheus metrics on `:9091`.
 
 ## Routine deploy
 
+### Archive hardening deployment hold
```
erskingardner: **Blocking.** This hold turns the PR's blockers into a standing rule for the runbook. The routine deploy below runs `git pull origin master`, so once this merges, every later deploy of any change either ships the fail-closed code or waits on these gates. And `master` would hold code the repo says not to run. The text also names this PR ("the archive-admission hardening"), so it goes stale as soon as the gates are met. The draft PR is already the right place for this hold. Please remove this section (and the "Do not deploy this change until…" sentence in `docs/archive_failure_recovery.md`) and merge only once the lifecycle, recovery, and notification gaps are fixed in code or config.

docs/archive_failure_recovery.md
```diff
@@ -0,0 +1,66 @@
...
+condition. This is intentionally conservative after a process/machine crash.
+
+Buffered bytes may flush when a failed writer is dropped, but the file is not
+promoted to a sealed archive segment. Preserve it unchanged for inspection.
+Automatic compression runs only after the authoritative uncompressed segment
+has been sealed; compression fallback is distinct from uncertain frame I/O.
+
+## Operator recovery gate
+
+This is a recovery **acceptance checklist**, not an executable repair procedure.
```
erskingardner: A policy doc that describes itself as a checklist "not an executable repair procedure" confirms the earlier finding: startup refuses `.open` files, and the repo has no path to recover them. Rewording the doc does not fix that. Either add the tested procedure, with concrete commands for each marker and `.open` case (extending `docs/archive_recovery.md`), or narrow the startup check so a crash-orphan `.open` does not need one. For example, refuse to start only when the recovery marker exists.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-23T14:15:11Z` |
| Commit reviewed | `b8b560b1a2f1c62e8ebc121e996f1af203125ce2` |

This re-reviews the PR after b8b560b. That commit changes only docs and one doc comment. No Rust behavior changed since fd8e101, so the earlier blocking findings still apply:

1. Any `.notepack.open` file or leftover marker stops startup. A routine unclean exit (OOM, SIGKILL at the stop timeout, reboot, or SIGKILL during a seal, when the per-seal marker is on disk) then crash-loops under `Restart=always`, and no tested recovery procedure exists. The same check also applies to `backfill-jsonl` and `backfill-proto`, so an interrupted backfill cannot resume without manual recovery.
2. On the live path, a dedupe `Err` now ends `run_async`, and the event is still lost.
3. After the writer latches, the live handler keeps reserving, dropping, and counting events as novel.

The author's comment lists these as open blockers, and I agree.

**New in b8b560b (inline):** the fix for undeployable code is to keep the PR unmerged, not to add a deploy hold to the runbook on `master`. Routine deploys run `git pull origin master`, so after this merges, every later deploy either ships this code or stays blocked. Please drop the hold and the "do not deploy this change" text from the policy doc, and fix the blockers in code.

**Non-blocking, not raised before:** `validate_archive_event` in `relay.rs` runs a Schnorr verification on every delivery before dedupe, including the many duplicates from overlapping relays. The SDK cache used to skip those. If you want to limit the CPU cost, verify only events that pass `reserve` (the drop guard already releases the claim on failure), plus kind 10002 before discovery. A forged copy of an ID that is already archived is dropped as a duplicate either way. The earlier non-blocking notes also still stand: the admission-gate mutex is held across the full seal, `write_reserved` is redundant, `admission.rs` could be inlined, the session-note docs, the missing Alertmanager target, and the `socket2` lock downgrade.

CI: rustfmt, clippy, and doctests pass. The three test shards were still running when I posted this. I ran nothing locally.

crates/pensieve-ingest/src/main.rs
```diff
@@ -1018,8 +1017,12 @@ async fn main() -> Result<()> {
                             }
                         };
 
+                        // Packing failures release the claim. Once the writer owns
+                        // bytes, an error may mean partial admission: retain the claim
+                        // until writer recovery or durable sealing resolves it.
                         // Write to segment - must succeed for event to be recorded in sync state
-                        if let Err(e) = negentropy_segment_writer.write(packed) {
+                        if let Err(e) = negentropy_segment_writer.write_reserved(packed, admission)
```
erskingardner: This comment says a write error is re-fetched next cycle. That is true only when the claim was released.

`write_reserved` calls `retain()` before the frame write. A failure after that leaves the id in the in-memory pending set. On the next cycle `reserve` returns `None`, this handler takes the duplicate branch and returns `Ok(true)`, and `run_periodic` skips sync-state because the id is not `Archived`. The bytes are never retried for the life of the process.

Once `archive_operation` has latched, later calls return before the closure runs, so those claims are dropped. They do come back, fail again immediately, and `run_periodic` keeps walking the batch. Negentropy then fails every event while the live handler also stays up. Stopping intake has to happen without exiting into `segment_writer.seal()?`, which is what `Restart=always` turns into the startup loop.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-23T14:18:38Z` |
| Commit reviewed | `b8b560b1a2f1c62e8ebc121e996f1af203125ce2` |
| Model | `Grok 4.7` |

`b8b560b` updates the recovery note and adds a RUNBOOK hold. The runtime on this SHA is the same as `fd8e101`. The hold records the gap. The process lifecycle is still the one in the existing threads, so this stays blocked.

**Blocking**
1. Startup still refuses every `.notepack.open` and every recovery marker. An unclean stop leaves an open segment, and `Restart=always` / `RestartSec=10` reopens RocksDB on each attempt. That path exits in `SegmentWriter::new`, after the metrics server is already up, and never sets `archive_recovery_required`. `up==0` for 2m does not stay true while each restart spends its time opening RocksDB.
2. A live dedupe error still returns `Err`, and `RelaySource::run_async` still breaks the subscription loop on that error. One RocksDB read error ends live ingestion.
3. A latched archive write still logs and continues on the live path (`Ok(true)`). `record_event` has already counted `admission.is_some()` as novel, and a latched `write_reserved` drops the new claim before any bytes are written. Novelty climbs for the rest of the outage.

The negentropy write-error comment claims the event is re-fetched next cycle. That holds only when the claim was released. See the inline note.

**Still open, non-blocking**
- `admission_gate` is held across seal fsync, `mark_archived`, and the watermark write.
- Passing `PendingAdmission` into the writer is redundant once every `write_inner` error latches the process.
- `admission.rs` wraps `event.verify()`; the tests belong next to dedupe and the writer.
- `docs/negentropy_audit_20260923.md` and `docs/negentropy_hardening_plan.md` are session notes in an admission PR.
- The new rules still have no Alertmanager receiver. The RUNBOOK now says that.
- `Cargo.lock` still downgrades `socket2` 0.6.1 to 0.5.10, and `Cargo.toml` floors `nostr` without flooring `nostr-relay-pool`.

The owned reservation, independent `Event::verify()`, and the `nostr` 0.44.7 / `nostr-relay-pool` 0.44.3 bump are the part worth keeping.

rustfmt, clippy, and doctests passed on this SHA. The three test shards were still running. Nothing was executed locally.

GitHub rejects a changes-request from the PR author, so this is a comment. Treat it as blocking.

erskingardner:
Round 2 (b8b560b): both independent reviewers confirm the runtime blockers remain. Keeping this draft and unmerged; the RUNBOOK hold is not a substitute for runtime recovery and will not be used as justification to merge. Resolving this needs the fail-closed recovery service lifecycle and tested recovery tooling, not a documentation-only change. The explicit user policy remains error/alert until recovery succeeds; silently ignoring orphan files or storage faults is not an accepted shortcut.

erskingardner:
Pushed eef6f0e after full just precommit passed (format, strict clippy, workspace tests and doctests). This also incorporates merged master/PR40. The live daemon now preflights recovery obligations before opening RocksDB, remains resident and observable on an archive fault, stops admission rather than continuing to count rejected events as novel, and rejects a live dedupe lookup failure without ending healthy subscriptions. Added /health/archive-admission on the metrics listener for the selected Uptime Kuma monitor: 503 before readiness, on recovery-required state, and during shutdown. Metrics bind failures now propagate. Regression tests cover preflight evidence preservation, resident recovery wait, health status transitions, and listener bind failure. No production deployment or Kuma configuration was performed. This is NOT yet merge-ready: tested complete-orphan/truncated/post-rename repair and reconciliation remain outstanding. Downstream replay must account for additive ClickHouse materialized views; ReplacingMergeTree alone does not make those counters replay-safe. Existing markers and evidence must not be cleared to bypass this gate.

crates/pensieve-ingest/src/pipeline/segment.rs
```diff
@@ -246,6 +255,37 @@ impl SegmentWriter {
...
+                    path.display()
+                )));
+            }
+        }
+        Ok(())
+    }
+
+    /// Whether archive admission has latched an uncertain write or seal fault.
+    pub fn recovery_required(&self) -> bool {
+        self.admission_gate.lock().is_some()
```
erskingardner: **Non-blocking.** `recovery_required()` locks `admission_gate`, and `archive_operation` holds that lock for the whole `seal_checkpointed()`: marker create and fsync, segment `sync_all` on the HDD, rename, directory fsyncs, `mark_archived` with `set_sync(true)`, and marker removal.

eef6f0e calls this on every live event, including the duplicates that make up most traffic (`main.rs` L1135). It also calls it on every negentropy event and from a tokio worker every 100ms (`main.rs` L805). So during each seal, all live intake parks behind the fsyncs, and so does one runtime worker. Before, only novel writes waited.

A latch is enough here:

```rust
latched: AtomicBool,
// in archive_operation, wherever `*failure = Some(..)` is set:
self.latched.store(true, Ordering::SeqCst);

pub fn recovery_required(&self) -> bool {
    self.latched.load(Ordering::SeqCst)
}
```

Keeping the gate across the whole seal is still the larger issue (see the earlier thread on L509), but this change fixes the new hot-path cost.

crates/pensieve-ingest/src/main.rs
```diff
@@ -774,7 +792,27 @@ async fn main() -> Result<()> {
...
-    let relay_source = RelaySource::with_manager(relay_config, Arc::clone(&relay_manager));
+    let relay_source = Arc::new(RelaySource::with_manager(
+        relay_config,
+        Arc::clone(&relay_manager),
+    ));
+    let fault_writer = Arc::clone(&segment_writer);
+    let fault_source = Arc::clone(&relay_source);
+    let fault_running = Arc::clone(&running);
+    let fault_ready = Arc::clone(&archive_ready);
+    let archive_fault_watch = tokio::spawn(async move {
```
erskingardner: **Nit.** This watch task and the per-event check at L1135 do the same job. After any latch, the handler returns `Ok(false)` on the next event, and the shutdown path at L1239 already clears readiness. The gauges are set by `archive_operation` and by `wait_for_archive_recovery`.

Once `recovery_required()` is an atomic load, you can delete this task (plus its abort/await at L1241-1242) and the duplicate `handler_running`/`archive_ready` stores in the write-error branch at L1188-1189. That removes about 20 lines. The only thing lost is prompt 503s when no relay traffic is arriving at all, which doesn't happen with this relay count.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-23T15:23:20Z` |
| Commit reviewed | `2178719b495da804676102d3bfbd215f6404b803` |

This round re-reviews eef6f0e and 2178719. The three runtime blockers from the earlier rounds are fixed in code:

1. **Crash loop.** The recovery check now runs before RocksDB opens. A marker or `.open` file keeps the process running with `archive_recovery_required=1` and the health endpoint returning 503. Because the process never exits, `Restart=always` has nothing to restart. The ctrlc `termination` feature means a SIGTERM stop still seals the final segment, so only a crash (OOM, SIGKILL, power loss, panic) leaves recovery work behind.
2. **Live dedupe error.** It now rejects only the one event and increments a counter. The subscription keeps running.
3. **Novelty inflation after the latch.** `record_event` now runs only after a successful write or on a duplicate. The handler stops on the first latched event.

The RUNBOOK "do not deploy" hold is now a list of deploy prerequisites, which is right. Manual, incident-specific recovery is an explicit operator decision, so I am not treating it as a blocker. The cost stays the same: any crash becomes a paged incident.

**Non-blocking (inline):**
- `recovery_required()` takes the same mutex that a seal holds across all of its fsyncs. The handler now calls it on every live event, duplicates included, and so do the negentropy handler and a tokio worker polling every 100ms. Use an `AtomicBool` latch.
- The fault-watch task and the per-event check do the same job. Keep one.

**Still open from earlier rounds (non-blocking; I won't repeat them inline):**
- `docs/negentropy_audit_20260923.md` and `docs/negentropy_hardening_plan.md` are session notes, about 275 lines.
- The `socket2` 0.6.1 to 0.5.10 downgrade in `Cargo.lock` is unrelated to this PR.
- `nostr-relay-pool` has no version floor in `Cargo.toml`.
- `admission.rs` is a one-line wrapper around `event.verify()`.
- `write_reserved` is redundant.
- The admission gate is held across the whole seal.

Each of these is a quick net-deletion cleanup.

**CI:** rustfmt, clippy, and doctests pass on this SHA. The three test shards were still pending at review time, so merge only once they are green. I ran nothing locally.

GitHub does not accept an approval from the PR author, so I filed this as a comment. The recommended action is **Merge**.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-23T15:29:11Z` |
| Commit reviewed | `2178719b495da804676102d3bfbd215f6404b803` |
| Model | `Grok 4.7` |

`eef6f0e` fixes the failure mode from the earlier reviews: recovery preflight runs before RocksDB, the process stays resident instead of exiting into `Restart=always`, a live dedupe error no longer tears down subscriptions, and novel relay counts are recorded only after a successful write. `/health/archive-admission` matches the Uptime Kuma plan. Remote CI on this SHA passed (rustfmt, clippy, doctests, and all three test shards).

The startup gate is still wider than the fault it is there to catch. A crash-orphan `segment-*.notepack.open` is the steady state after SIGKILL, OOM, or a stop that times out. This writer already leaves that file untouched and opens the next segment number. Refusing to start, while the policy doc says a procedure for that case is out of scope, leaves admission down with no exit. Gate startup on `segment.recovery-required`. Quarantine a crash-orphan `.open` without sealing it: those IDs were never marked `Archived`.

Still open, and not blocking on their own: the admission mutex is held for the whole seal, the session-note docs are still in this PR, and `Cargo.lock` still moves `hyper-util` / `quinn` / `quinn-udp` from socket2 0.6.1 to 0.5.10 while `Cargo.toml` does not floor `nostr-relay-pool`.

crates/pensieve-ingest/src/pipeline/segment.rs
```diff
@@ -246,6 +255,37 @@ impl SegmentWriter {
...
+                marker.display()
+            )));
+        }
+        if !output_dir.try_exists()? {
+            return Ok(());
+        }
+        for entry in fs::read_dir(output_dir)? {
+            let path = entry?.path();
+            let name = path.file_name().unwrap_or_default().to_string_lossy();
+            if name.starts_with(&format!("{prefix}-")) && name.ends_with(".notepack.open") {
```
erskingardner: **Blocking.** The resident wait keeps the process up. This condition still stops admission after every unclean exit.

During ingest the current file is `segment-*.notepack.open`. SIGKILL, OOM, or a stop that hits the unit's stop timeout (this unit does not set `TimeoutStopSec`) leaves that file behind. `find_next_segment_number` already counts `.notepack.open`, and the next `File::create` uses the following number. This writer does not append to the orphan or seal it.

Those IDs are not `Archived`. The pending set is memory-only and died with the process. Quarantine the file out of the active `segment-*.notepack.open` namespace and start. Do not seal it. Sealing it after the same events are re-fetched is what double-counts additive ClickHouse views.

`segment.recovery-required` is the startup gate that matches the hazard: it is written only when a seal may already have renamed the segment. Please refuse startup for that marker, and treat a crash-orphan `.open` as evidence to preserve beside the active namespace.

docs/archive_failure_recovery.md
```diff
@@ -0,0 +1,94 @@
...
+
+Buffered bytes may flush when a failed writer is dropped, but the file is not
+promoted to a sealed archive segment. Preserve it unchanged for inspection.
+Automatic compression runs only after the authoritative uncompressed segment
+has been sealed; compression fallback is distinct from uncertain frame I/O.
+
+## Incident-specific operator recovery
+
+Operator decision, 2026-09-23: recovery is manual and specific to the incident.
+A universal repair tool or prewritten procedure for every failure is not a merge
```
erskingardner: **Blocking.** This is the sentence that leaves the `.open` startup check with no exit. The check turns a crash-orphan open segment into an incident that stays open until a repair finishes. The following paragraphs say a procedure for that case is out of scope, and that the current tools cannot recover a complete `.open` orphan.

"No universal repair tool" can stand. The common case still needs a tested disposition: quarantine the orphan without sealing it, because those IDs were never marked `Archived`, then restart. A marker-only startup gate covers the post-rename window and does not require that tool.

crates/pensieve-ingest/src/pipeline/admission.rs
```diff
@@ -0,0 +1,106 @@
...
+        assert!(result.is_err());
+        assert!(dedupe.is_new(&id).unwrap());
+        dedupe.reserve(&id).unwrap().unwrap().retain();
+        assert!(!dedupe.is_new(&id).unwrap());
+        dedupe.mark_archived(std::iter::once(&id)).unwrap();
+        assert!(!dedupe.is_new(&id).unwrap());
+    }
+
+    #[test]
+    fn segment_open_failure_releases_claim_for_retry() {
```
erskingardner: Open failure goes through `archive_operation`, which latches the writer and fsyncs `segment.recovery-required` before this returns. The claim is released, and both a later `reserve` in this process and `SegmentWriter::new` on restart fail. Please name the test for that latched outcome. The current name says the release exists so a retry can proceed.

erskingardner:
Round disposition: both reviewer processes exited cleanly (0). Claude recommended Merge on 2178719; Grok requested changes to the orphan-file startup policy. Remote CI passed on that reviewed head. Pushed 86ae883 after full just precommit: recovery_required now observes an AtomicBool instead of waiting on the archive admission mutex; a regression holds the mutation lock while the observer completes, then verifies the latch stays true after failure. Renamed and strengthened the open-failure test to distinguish released reservation from a blocked writer.

Retaining the watcher as well as per-event rejection: the watcher withdraws health even with no incoming relay traffic; the per-event check rejects immediately at the admission boundary.

Not implementing automatic orphan quarantine/resumption. The operator explicitly selected incident-specific manual recovery and remaining blocked until investigation/repair. Marker absence cannot distinguish a benign crash from a failed frame write when disk exhaustion prevented marker persistence; the open file is the fallback evidence in that case. Manual recovery is intentionally not automated, not impossible. This is a deliberate safety/availability tradeoff, not an unimplemented universal-repair feature. The PR stays unmerged because the requested unanimous merge recommendation has not been obtained. No production changes or monitor configuration were made.

Issue #39, opened 2026-09-23
Plan spare-capacity historical negentropy exploration


erskingardner:
# Follow-up: spare-capacity historical negentropy exploration

Deferred until the isolated negentropy worker, durable accounting and production
recovery/soak gates are proven. This is not part of the initial reliability rollout.

## Goal

Gradually recover older events from useful relays without starving current recovery,
retained unfinished jobs, live ingestion or archive publication. Reuse the bounded
worker and job ledger; do not introduce a graph database or SDK fork.

## Required foundation

- Initial reconciliation covers a rolling 14 days; unfinished gaps do not age out.
- Lightweight history records relay, inclusive time interval, checked-at time,
  outcome/reason, received events, durably novel events, duplicates, bytes and runtime.
- Pending admission is not durable novelty; failures are not empty results.
- Successful zero-new-event reconciliation means nothing missing was returned,
  not that the relay stores no events. Observations are time-specific, not proof
  of complete global history or the relay's founding date.

## Proposed exploration policy

1. Reserve capacity for recent recovery and fair retries of old unfinished work.
2. Give historical exploration a small explicit budget, gated by CPU, disk reserve,
   ingestion health, queue depth and archive-publication lag. Pause it first under load.
3. Start with an explicit trusted relay set and expand only after measured evidence.
4. Explore progressively older bounded intervals with deterministic split/retry rules.
5. Repeated successful zero-novel checks lower recheck frequency; they do not establish
   a permanent historical cutoff. Periodic sparse probes allow for later imports.
6. Persist failure backoff separately from low-yield history. Minimum-size intervals
   that still exceed limits remain visible and actionable, not silently discarded.
7. Keep storage bounded: compact completed observations; preserve unresolved jobs
   and summary evidence. No unlimited payload or attempt-log retention.

## Acceptance criteria

- Recent jobs and oldest unresolved gaps make progress while exploration runs.
- Exploration stops promptly when its resource/health budget is unavailable.
- Restart preserves progress and failures; failed jobs never become empty coverage.
- A relay that later gains older events is eventually probed again.
- Tests distinguish empty relay, already-known inventory, novel historical events,
   connection failure, partial transfer, resource exhaustion and late arrival.
- Report recovery yield and resource cost before expanding scope. A simple
   relay-by-time report may follow; exact relay inventories/heatmaps are not required.

Decide initial exploration budget, historical horizon, revisit cadence and relay
expansion policy when this follow-up starts, using the accumulated accounting.

PR #42, opened 2026-09-23
Bound negentropy relay lifecycles and local inventory

erskingardner:
Draft follow-on to archive admission hardening. Isolates per-relay deadlines and cleanup, bounded streaming admission, and local inventory reads and cold-start seeding. NOT ready to merge or deploy: the inventory cap can reject the existing 14-day request until resumable window scheduling exists; SDK remote inventory allocation is not process-isolated yet. The approved isolated-worker redesign is separate future work; historical exploration is tracked in #39. Review this foundation now, but keep this PR draft until those integration gates are satisfied.


crates/pensieve-ingest/src/sync/negentropy.rs
```diff
@@ -67,6 +79,10 @@ impl Default for NegentropySyncConfig {
             interval: Duration::from_secs(1800), // 30 minutes
             lookback: Duration::from_secs(14 * 24 * 3600), // 14 days
             protocol_timeout: Duration::from_secs(900), // 15 min for full sync
+            max_concurrent_relays: 3,
+            event_queue_capacity: 128,
+            max_event_bytes: 1024 * 1024,
+            max_inventory_items: 250_000,
```
erskingardner: **Blocking.** The PR body acknowledges this; I'm restating it here so it stays attached to the code. `seed_from_clickhouse` seeds *every* event in the lookback (`WHERE created_at >= now() - INTERVAL {lookback} DAY`), and the default lookback is 14 days. Production inventory for that window is far above 250,000 IDs, so `get_items_range` fails for every relay on every cycle and negentropy recovers nothing. Because this PR targets `codex/archive-admission-hardening`, merging it puts that branch in the same state.

Either land the window scheduler in this PR, or keep the cap inert until the scheduler exists (for example, default `usize::MAX` and flip it in the scheduler PR).

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-23T13:51:31Z` |
| Commit reviewed | `dda3b2c0ed26dec7399123268425026b4d6cbe65` |

**Summary.** The direction is right. This PR replaces the unbounded `Vec<Event>` collector and the pool-wide timeout with per-relay single-`Relay` futures that own their cleanup, streams admission through a bounded queue, and streams the ClickHouse seed. All of these are real improvements. As the body says, the PR isn't mergeable yet, and two of the new limits make negentropy fail under normal production load even before deploy.

**Blocking**
1. **The inventory cap guarantees failure.** The cap is 250k IDs, but the lookback is 14 days and startup seeds that entire window from ClickHouse, so every relay fails every cycle. The same would be true on the `codex/archive-admission-hardening` base.
2. **The lossy queue turns normal bursts into failures.** The queue is 128 slots shared by 3 relays, using `try_send`, while the SDK fetches in REQ batches of 100 per relay. Any consumer stall becomes a relay failure, and the SDK logs one ERROR with the full event JSON per dropped event. Per-relay `send().await` is simpler and loses nothing, because each worker already owns its own client and deadline.

**Non-blocking**
- Removing pruning isn't needed for correctness. It makes sync-state grow without bound and drops the gauge the plan says to monitor.
- `admission_failed` duplicates the existing `remote ⊆ received` check (~15 lines to delete). The adapter's size cap duplicates the SDK's `RelayEventLimits::max_size`.
- One unservable or expired ID fails a relay forever. Because `negentropy_last_sync_unix` only advances when every relay succeeds, the freshness gauge can freeze.
- `run_periodic` now returns on the first handler error. That ends negentropy for the rest of the process's life, with only a log line and `negentropy_sync_errors_total` as signal. This is consistent with #41's fail-closed stance, but the operations-slice alert or supervisor needs to exist before deploy.
- Dead code and error labelling: the `timestamp > until` branch, `get_items_since` (now test-only), and `Error::Config` used for runtime overflow and cancellation.

**Validation:** rustfmt passed. clippy, tests and doctests were still running at review time. No local execution was done.

_Posted as a comment because GitHub does not allow requesting changes on your own PR._

crates/pensieve-ingest/src/sync/negentropy.rs
```diff
@@ -136,19 +219,25 @@ impl NostrDatabase for SyncStateAdapter {
...
+            // Never block an SDK notification loop on archive backpressure.
+            // A rejected event makes this relay incomplete even if SDK sync returns Ok.
+            let result = (|| {
+                if event.as_json().len() > self.max_event_bytes {
+                    return Err(DatabaseError::backend(std::io::Error::other(
+                        "negentropy event exceeds admission size cap",
+                    )));
+                }
+                crate::pipeline::validate_archive_event(event).map_err(DatabaseError::backend)?;
+                self.event_sender.try_send(event.clone()).map_err(|_| {
```
erskingardner: **Blocking.** With `try_send` into a 128-slot queue shared by 3 relays, ordinary bursts become relay failures. nostr-relay-pool 0.44.3 fetches missing IDs in REQ batches of `NEGENTROPY_BATCH_SIZE_DOWN = 100` (`relay/constants.rs`), so about 300 events can be in flight at once. Meanwhile the single consumer runs a synchronous handler that can stall on segment rotation/fsync or on the segment-writer lock it shares with live ingest. Every drop also reaches the SDK's `tracing::error!(msg = %msg, ...)` in `handle_relay_message`, which logs the full raw event JSON at ERROR once per dropped event.

Each worker owns its own `Client`, so blocking this relay's notification loop only stalls this relay, and `protocol_timeout` still bounds it. `self.event_sender.send(event.clone()).await` (or `send_timeout`) gives lossless per-relay backpressure and removes the "backpressured" failure class entirely. If you keep fail-fast, capacity needs to be at least `max_concurrent_relays * 100`.

crates/pensieve-ingest/src/sync/negentropy.rs
```diff
@@ -291,324 +400,237 @@ impl NegentropySyncer {
...
+                .sync(filter, &opts)
+                .await
+                .map_err(|e| compact_error(&e))?;
+            if !output.remote.is_subset(&output.received) {
+                return Err("relay reconciliation left remote events unfetched".to_string());
+            }
+            tracing::info!(relay_url = %url, phase = "disconnect", "Negentropy relay phase");
+            relay.0.disconnect();
+            client.shutdown().await;
+            if failed.load(Ordering::SeqCst) {
```
erskingardner: `admission_failed` duplicates the `remote.is_subset(&output.received)` check at L448. When `save_event` returns `Err`, `handle_raw_relay_message` returns `Err`. `handle_relay_message` then never broadcasts `RelayNotification::Message`, so the ID never enters `output.received` (inner.rs ~996-1070 and ~2023). Every rejected, oversized or backpressured event therefore already fails the subset check.

You can delete the field, the `Arc`, the closure-plus-`store` in `save_event` and this branch (~15 lines). `adapter_backpressure_and_oversize_fail_closed` would then assert only on the `save_event` result.

crates/pensieve-ingest/src/sync/negentropy.rs
```diff
@@ -291,324 +400,237 @@ impl NegentropySyncer {
...
+            relay.0.wait_for_connection(Duration::from_secs(10)).await;
+            tracing::info!(relay_url = %url, phase = "reconcile", "Negentropy relay phase");
+            let opts = SyncOptions::default().direction(self.config.direction);
+            // Single relay API: no pool-wide join_all and no misleading Output.success.
+            let output = relay
+                .0
+                .sync(filter, &opts)
+                .await
+                .map_err(|e| compact_error(&e))?;
+            if !output.remote.is_subset(&output.received) {
```
erskingardner: Non-blocking. "Not complete" is the correct answer here, but one unservable ID now fails the relay on every cycle. Examples: an event the relay still lists in its negentropy set but that the SDK rejects (NIP-40 expired → `Error::EventExpired`), or one deleted between reconcile and REQ. Combined with `negentropy_last_sync_unix` advancing only on all-relay success (L518) and catalog-expanded targets, the freshness gauge can freeze permanently over a few stale IDs.

Consider carrying `missing: usize` (`remote.difference(&received).count()`) in `RelaySyncResult` instead of only an error string, so operators can tell "relay down" apart from "3 IDs unservable".

crates/pensieve-ingest/src/sync/negentropy.rs
```diff
@@ -291,324 +400,237 @@ impl NegentropySyncer {
...
-                tokio::time::sleep(Duration::from_secs(1)).await;
+            };
+            tracing::info!(
+                received = stats.events_received,
+                completed = stats.relays_responded,
+                failed = stats.relays_errored,
+                elapsed_secs = stats.duration.as_secs_f64(),
+                "Negentropy cycle finished"
+            );
+            // Inventory pruning/coverage policy is a separate slice. Do not prune
```
erskingardner: Removing pruning isn't needed for correctness, and it makes sync-state grow without bound. `prune_before(now - lookback)` only deletes keys older than the next cycle's `since`, and `get_items_range` never reads those keys, so pruning can't hurt a partial or failed cycle. The only consumer of the old keys would be the not-yet-designed resumable-window scheduler.

This also drops the `negentropy_sync_state_items` gauge, which is the metric the plan doc tells operators to watch ("monitor sync-state disk use"). Suggest restoring the prune + gauge block until the coverage slice actually needs the old entries.

crates/pensieve-ingest/src/sync/negentropy.rs
```diff
@@ -136,19 +219,25 @@ impl NostrDatabase for SyncStateAdapter {
...
-            self.event_sender.send(event.clone()).map_err(|_| {
-                DatabaseError::backend(std::io::Error::new(
-                    std::io::ErrorKind::BrokenPipe,
-                    "negentropy admission channel closed",
-                ))
-            })?;
+            // Never block an SDK notification loop on archive backpressure.
+            // A rejected event makes this relay incomplete even if SDK sync returns Ok.
+            let result = (|| {
+                if event.as_json().len() > self.max_event_bytes {
```
erskingardner: Nit: the SDK already implements this. `RelayEventLimits::max_size` (via `RelayOptions::limits` + `add_relay_with_opts`) checks the same normalized-JSON size in `handle_event_msg`, *before* signature verification, whereas this re-serializes every event after verification. The raw message cap is already 5 MiB by default.

crates/pensieve-ingest/src/sync/state.rs
```diff
@@ -131,6 +131,51 @@ impl SyncStateDb {
...
+            IteratorMode::From(&start_key, rocksdb::Direction::Forward),
+            options,
+        );
+        let mut items = Vec::with_capacity(max_items.min(1024));
+        for item in iter {
+            let (key, _) = item?;
+            let (id, timestamp) = Self::parse_key(&key).ok_or_else(|| {
+                crate::Error::Database("malformed sync inventory key".to_string())
+            })?;
+            if timestamp > until {
```
erskingardner: Nit: dead branch. When `until < u64::MAX`, the iterate upper bound already stops at `until + 1`. When `until == u64::MAX`, no timestamp can exceed it. Also:
- `get_items_since` is now used by only one test (negentropy.rs L995). Switch that test to `get_items_range` and delete it.
- The `Error::Config` returned here for overflow (and in `sync_once`/`drive_workers` for cancellation and handler stop) labels runtime conditions as configuration errors.

crates/pensieve-ingest/src/sync/negentropy.rs
```diff
@@ -291,324 +400,237 @@ impl NegentropySyncer {
...
+        loop {
+            if *stop.borrow() {
+                break;
+            }
+            let stats = match self.sync_once(&mut event_handler).await {
+                Ok(stats) => stats,
+                Err(_) if *stop.borrow() => break,
+                Err(error) => {
                     counter!("negentropy_sync_errors_total").increment(1);
+                    return Err(error);
```
erskingardner: `return Err(error)` ends the spawned negentropy task. `main` only logs `negentropy sync task failed` and leaves live ingestion running; nothing calls `run_periodic` again.

`stop()` is fine: the handler's `Ok(false)` latches the watch channel and this arm breaks. Every other error is fatal for the rest of the process. That includes a dedupe `reserve` failure, a notepack pack failure, `get_status` returning `Err`, and `sync_state.record` returning `Err`. None of those latch the archive writer. `archive_operation` already latches real segment I/O failures process-wide, so exiting here does not add that protection.

The handler in `main` still says a pack failure is not recorded and will be re-fetched on the next cycle. That was true when this loop counted the failure and continued. It is not true now: one bad event disables reconciliation until the ingester process restarts.

Fail that event closed (do not `record` it) and keep the cycle and the periodic loop running unless `stop()` was requested.

crates/pensieve-ingest/src/sync/negentropy.rs
erskingardner: This `try_send` fails a healthy relay once admission falls behind a single SDK down-batch. `nostr-relay-pool` 0.44.3 keeps one down REQ in flight of `NEGENTROPY_BATCH_SIZE_DOWN` (100) ids, then sends the next 100 after EOSE. `save_event` runs on the connection task, which can deliver that next batch while this task is still draining the previous one. Peak occupancy is about 200 events. The default `event_queue_capacity` is 128, so the second batch takes the `Err` path.

The pool logs that error and continues (`handle_relay_message` swallows `save_event` failures). The id never enters `Reconciliation.received`, and `remote.is_subset(received)` then marks the whole relay incomplete. Those events were valid; the queue was just shorter than the protocol window.

Size the shared queue for at least two down-batches, and add a regression that admits two full 100-id batches without setting `admission_failed`. `try_send` can stay non-blocking.

crates/pensieve-ingest/src/sync/negentropy.rs
```diff
@@ -661,19 +683,16 @@ pub async fn seed_from_clickhouse(
...
         .map_err(crate::Error::ClickHouse)?;
 
-    tracing::info!("Fetched {} events from ClickHouse for seeding", rows.len());
-
     // Convert and insert into sync state
     let mut count = 0usize;
     let mut batch = Vec::with_capacity(10_000);
 
-    for row in rows {
+    while let Some(row) = rows.next().await.map_err(crate::Error::ClickHouse)? {
```
erskingardner: These rows are written in 10k batches before the stream finishes. `main` only seeds when `is_empty()` (`approximate_count() < 100`), so a ClickHouse error after the first batch is not retried on restart. The missing ids are re-downloaded rather than dropped, but the log reads like a retryable seed failure. Commit the seed only after the stream completes, or persist a cursor a later start can resume. The plan note that a partial stream is not coverage is right; the `is_empty` gate makes that partial stream sticky.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-23T13:56:01Z` |
| Commit reviewed | `dda3b2c0ed26dec7399123268425026b4d6cbe65` |
| Model | `Grok 4.7` |

The per-relay deadline, fail-closed inventory cap, and durable-before-record gate are the right foundation, and the draft gates you already called out (14-day cap vs window scheduler, unbounded SDK ID sets, no process isolation) should stay closed. Two behaviors in this slice will stop a healthy reconciliation before those gates matter.

`rustfmt` had passed at review time. Clippy and the test shards were still running, so this review does not treat CI as green.

Serious:
- A retryable admission error ends the negentropy task for the life of the process. Live ingestion keeps going, and nothing respawns the task.
- The shared 128-slot `try_send` queue is smaller than two SDK down-batches (100 ids each). A fast relay fails closed while valid events are still in flight.

Also, a mid-stream seed error is sticky: batches are committed before the stream finishes, and `main` will not seed again once `is_empty()` is false.

erskingardner:
Review response (one address pass): e406a5a replaces lossy try_send with bounded asynchronous backpressure and adds a 200-event burst regression. Interrupted streamed seeds now retain a durable retry marker across reopen; it is cleared only after a complete flushed seed, with a restart regression. All 19 sync tests and full just precommit passed. This remains draft: finite-window scheduling and the admission-error supervisor/recovery lifecycle are not implemented. The inventory cap remains fail-closed; it was not weakened to bypass the missing scheduler. No deployment or merge is authorized by these test results.

crates/pensieve-ingest/src/sync/state.rs
```diff
@@ -77,7 +78,30 @@ impl SyncStateDb {
...
 
-        Ok(Self { db })
+        Ok(Self {
+            db,
+            seed_marker: path.join("seed-in-progress"),
+        })
+    }
+
+    /// Retry interrupted seeds even when earlier streamed batches are present.
+    pub fn needs_seed(&self) -> Result<bool> {
```
erskingardner: This marker addresses an efficiency cost, not a correctness problem, and it only helps in a narrow window.

- A partial seed is a *subset* of archived IDs, so it can never advertise an event we lack. The worst case is that the first cycle re-requests the missing IDs. Dedupe then reports them `Archived`, and `admit_event` records them, so the inventory repairs itself.
- `needs_seed` is only consulted at process start (`main.rs:704`). The marker therefore saves work only when the process restarts *before* that first repairing cycle completes.

Costs: a filesystem side-channel beside RocksDB, parent-directory fsyncs duplicated in `begin_seed` and `complete_seed`, a test placed above the module's `use tempfile::TempDir;`, and `needs_seed()?` in `main.rs`. That last one turns a `try_exists` error into an ingest startup failure, where the old code was `unwrap_or(true)`. The `else` branch at `main.rs:730` also still logs "Sync state is empty" when the reason is the marker.

Suggest reverting to `is_empty().unwrap_or(true)` and deleting `needs_seed`, `begin_seed`, `complete_seed` and the test. If a partial seed matters operationally, a counter or log on the error path is enough.

crates/pensieve-ingest/src/sync/negentropy.rs
```diff
@@ -291,324 +401,237 @@ impl NegentropySyncer {
...
+                .sync(filter, &opts)
+                .await
+                .map_err(|e| compact_error(&e))?;
+            if !output.remote.is_subset(&output.received) {
+                return Err("relay reconciliation left remote events unfetched".to_string());
+            }
+            tracing::info!(relay_url = %url, phase = "disconnect", "Negentropy relay phase");
+            relay.0.disconnect();
+            client.shutdown().await;
+            if failed.load(Ordering::SeqCst) {
```
erskingardner: With `send().await`, this flag can only be set by oversize, validation failure, or a closed channel. Every one of those makes `save_event` return `Err`. In nostr-relay-pool 0.44.3 (`inner.rs:1234`), an `Err` from `save_event` means `RelayMessage::Event` is never produced, so the ID never reaches `output.received`. The `remote.is_subset(&received)` check above already fails this relay.

The message also still says "backpressured", which can no longer happen. Suggest deleting `admission_failed` (the field, this block and the `store`), for a net deletion of about 15 lines.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-23T14:16:06Z` |
| Commit reviewed | `e406a5abfbf6c12a766594572b81e82ede139ba4` |

**Re-review of e406a5a.** This pass covers the backpressure and seed-retry changes. The earlier findings still apply unless they are listed as resolved below.

**Resolved**
- Lossy queue: `send().await` fixes it. Backpressure stays inside each relay's connection task. Cancellation and the end-of-cycle `receiver.close()` both release blocked senders, and nothing can deadlock, because the consumer runs in the same `select!` as the workers. The 200-event regression test is a good one.

**The seed-marker premise is weak (net deletion recommended)**
- A partial seed only costs efficiency, not correctness, and `main.rs` already says so ("first negentropy sync may request more events than necessary"). A subset inventory never claims an event we lack. The next cycle re-fetches the missing IDs, and dedupe reports them `Archived`, so `admit_event` records them. The inventory repairs itself in one cycle.
- The marker is only checked at process start. It therefore helps only when the process restarts before that first cycle finishes. In exchange, it adds a filesystem side-channel beside RocksDB, two directory fsyncs, a test, and a new way for ingest to fail at startup. See inline.

**Still open from earlier reviews (acknowledged draft gates, unchanged by this commit)**
- The 250k inventory cap against a complete 14-day seed still makes every relay fail every cycle. A *successful* seed from the new marker path lands in exactly that state.
- `run_periodic` still returns on the first admission error, and nothing restarts it.
- Automatic pruning is still removed, and nothing bounds sync-state growth yet.

**New, non-blocking**
- `admission_failed` is now fully redundant with `remote ⊆ received`, and its message mentions backpressure, which is gone.
- The plan doc still describes overflow as a failure.
- The backpressure comment names the wrong bound: the SDK's 55-second ping disconnects the relay long before the 900-second deadline would.

**Validation:** rustfmt passed. clippy, the three test shards and doctests were still pending at review time. No local execution was done.

crates/pensieve-ingest/src/sync/negentropy.rs
```diff
@@ -136,19 +219,26 @@ impl NostrDatabase for SyncStateAdapter {
...
-            //
-            // This ensures that if segment write fails, the event is NOT
-            // recorded in sync-state and will be re-fetched on the next cycle.
-            self.event_sender.send(event.clone()).map_err(|_| {
-                DatabaseError::backend(std::io::Error::new(
-                    std::io::ErrorKind::BrokenPipe,
-                    "negentropy admission channel closed",
-                ))
-            })?;
+            // Backpressure only this relay's connection task. Its outer lifecycle
```
erskingardner: Minor: the effective bound on a stalled consumer is the SDK's 55-second `PING_INTERVAL`, not the 900-second lifecycle deadline. `save_event` runs inline in `receiver_message_handler`, which is also the loop that reads Pongs. After about 55–110 seconds blocked here, `sender_message_handler` returns `NotRepliedToPing`, the relay disconnects, and `sync` returns `NotConnected`. That outcome is acceptable, because the relay fails closed. Please say so in the comment, so nobody tunes `protocol_timeout` expecting it to govern this case.

docs/negentropy_hardening_plan.md
```diff
@@ -17,10 +17,39 @@ See `negentropy_audit_20260923.md` for the original findings and limitations.
   are not yet verified.
 - [ ] Bounded independent relay workers: full-lifecycle deadline, idle deadline,
   structured child ownership, bounded streaming admission, per-relay results.
+  Lifecycle/queue slice implemented: three concurrent single-relay futures,
+  independent 900-second deadlines spanning setup through shutdown, synchronous
+  disconnect-on-drop, no detached collector, immediate streaming admission,
+  128-event shared queue and 1 MiB normalized event cap. Overflow/oversize/invalid
```
erskingardner: Stale after e406a5a: queue overflow now backpressures instead of failing the relay. Line 27 ("final verification is rerun after the added admission-failure regression before commit") is also process narration and will go stale. Suggest removing it.

crates/pensieve-ingest/src/sync/negentropy.rs
```diff
@@ -291,324 +401,237 @@ impl NegentropySyncer {
...
+        let durable = match &self.dedupe {
+            Some(dedupe) => matches!(
+                dedupe.get_status(event.id.as_bytes())?,
+                Some(EventStatus::Archived)
+            ),
+            None => true,
+        };
+        if durable {
+            self.sync_state
+                .record(event.id.as_bytes(), event.created_at.as_secs())?;
```
erskingardner: These two `?`s abort the cycle. `get_status` used to treat a read error as "not durable" and continue; `record` used to log and continue. Either failure now returns from `drive_workers`, and `run_periodic` does not run another cycle.

The durable-before-record rule is satisfied by skipping `record` when status is not `Archived` or the bookmark write fails. The event stays out of sync-state and is eligible again later. Handler `Err` is already the archive fail-closed path. A RocksDB read blip or a sync-state put should not take that same path until something restarts the task.

docs/negentropy_hardening_plan.md
```diff
@@ -17,10 +17,39 @@ See `negentropy_audit_20260923.md` for the original findings and limitations.
...
+  delivery makes a relay incomplete. Parent cancellation drops worker futures;
+  shutdown now awaits the periodic task after cancellation/abort. Tests cover
+  independent progress, timeout/panic/drop, backpressure, restart after abort,
+  and a silent loopback WebSocket relay. Full workspace precommit passed; final
+  verification is rerun after the added admission-failure regression before commit.
+  This is NOT an end-to-end memory or coverage bound: SDK reconciliation ID sets
+  and local inventory scans remain unbounded. Async deadlines cannot interrupt
```
erskingardner: This paragraph still describes the previous queue. A full queue now waits in `send().await`; it does not by itself mark the relay incomplete. Local scans are also capped: `get_items_range` errors above 250,000 ids. The next bullet states that cap. This sentence says the scans remain unbounded, which will send the next slice after the wrong bug.

Update this note so overflow means backpressure, and reserve "unbounded" for the SDK's remote ID sets.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-23T14:21:46Z` |
| Commit reviewed | `e406a5abfbf6c12a766594572b81e82ede139ba4` |
| Model | `Grok 4.7` |

The address pass fixes the two defects from the last review. `save_event` now waits on `send().await`, and the 200-event test drives a one-slot queue. Interrupted seeds keep `seed-in-progress` until `complete_seed` flushes and unlinks it, including across reopen.

Keep this draft. After a streamed 14-day seed succeeds, the 250k cap still fails every relay on every cycle. A handler error still returns from `run_periodic`, the spawned task only logs it, and nothing starts negentropy again. The handler comments in `main` still say that error is retried on the next cycle.

`rustfmt` passed. Clippy, doctests, and the three test shards were still running, so this review does not treat CI as green. No local execution.

erskingardner:
Round 2 (e406a5a): both reviewers confirm lossless bounded backpressure; Grok also confirms the partial-seed retry fix. Keeping the retry marker: it addresses restart efficiency, not a claim of recovery coverage, and does not replace the scheduler. This remains draft and unmerged pending finite-window scheduling, persistent unfinished-gap accounting, supervision after admission failures, and a bounded inventory retention policy. These are the already-planned implementation gates, not something another unchanged review can clear.

PR #43, opened 2026-09-23
Harden archive retention cleanup

agent-sv3n:
## Summary

- set the default and example local Notepack recovery window to 90 days
- recognize both `.notepack` and `.notepack.gz` sealed archives
- build a fresh Storage Box inventory and require an exact remote path-and-size match before deleting a local archive
- add retention validation, dry-run support, safer counters, and explicit cleanup reporting

## Verification

- `bash -n ops/scripts/sync-archive.sh`
- `just precommit`
- production dry-run and cleanup completed successfully before this PR; mismatched segment 6397 was preserved


ops/scripts/sync-archive.sh
```diff
@@ -63,22 +83,55 @@ log "Disk usage: ${DISK_USAGE}%"
...
-            ((CLEANED++))
+        local_size=$(stat -c '%s' "$file")
+        remote_size="${REMOTE_SIZES[$filename]:-}"
+
+        if [ -z "$remote_size" ]; then
+            log "Keeping (not present in remote inventory): $filename"
+            ((MISSING_REMOTE += 1))
+        elif [ "$local_size" != "$remote_size" ]; then
+            log "Keeping (size mismatch local=$local_size remote=$remote_size): $filename"
+            ((SIZE_MISMATCH += 1))
```
erskingardner: A mismatch here means the remote backup of this segment is truncated or wrong, and because Step 1 uses `--ignore-existing` it will never be re-uploaded. The local copy is kept, but the only signal is a log line, and the unit still exits 0, so nothing alerts and the bad remote copy stays unnoticed until it is needed.

Suggest exiting non-zero after the summary when `SIZE_MISMATCH > 0` (still after the loop, so other eligible files get cleaned) so `archive-sync.service` shows as failed:

```bash
(( SIZE_MISMATCH == 0 )) || { log "ERROR: $SIZE_MISMATCH remote size mismatch(es); re-upload required"; exit 1; }
```

Repairing 6397 itself (e.g. a one-off `rclone copyto` without `--ignore-existing`) can happen separately.

ops/scripts/sync-archive.sh
```diff
@@ -63,22 +83,55 @@ log "Disk usage: ${DISK_USAGE}%"
...
+    trap 'rm -f "$REMOTE_INVENTORY"' EXIT
+    rclone lsjson "$REMOTE_NAME:$REMOTE_PATH" \
+        --files-only \
+        --recursive \
+        > "$REMOTE_INVENTORY"
+
+    declare -A REMOTE_SIZES=()
+    while IFS=$'\t' read -r remote_path remote_size; do
+        REMOTE_SIZES["$remote_path"]="$remote_size"
+    done < <(jq -r '.[] | select(.IsDir == false) | [.Path, (.Size | tostring)] | @tsv' "$REMOTE_INVENTORY")
```
erskingardner: Nit: `select(.IsDir == false)` is redundant with `--files-only`. `--recursive` also does nothing useful: uploads are root-anchored (`+ /segment-*`) and the lookup key is `basename`, so nested entries can never match. It just makes a full walk of the remote tree on every cleanup run. Dropping `--recursive` and the `select` keeps the same behavior with less to read.

erskingardner (approved):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-23T14:21:07Z` |
| Commit reviewed | `6febd158707470b81a1db34e02d0c4c4afb6fbe5` |

Strictly safer than what it replaces: the old loop deleted on bare remote existence, only matched uncompressed `.notepack`, and `((CLEANED++))` from 0 returns status 1 under `set -e`, so the first deletion killed the script. The fresh single `lsjson` inventory with a path-and-size match, the `.notepack.gz` coverage, and the `+= 1` counters fix all three. Config validation is reasonable in a script that deletes things. The `rclone lsjson` failure aborts via `set -e` before any `rm`, so it fails closed.

Two non-blocking notes inline. The one I'd most like addressed: a size mismatch means the Storage Box copy is bad, and `--ignore-existing` guarantees it will never be repaired. Right now that only shows up as a log line with exit 0 (segment 6397 is in exactly that state).

PR #44, opened 2026-09-23
Pin production Compose runtime topology

agent-sv3n:
## Summary

- make the production `postgres-analytics` service manual-only and disable its automatic restart policy
- pin the production Compose default network to `172.18.0.0/16` with gateway `172.18.0.1`

## Verification

- `docker compose --env-file /etc/pensieve/pensieve.env -f ops/production/compose.yml config --quiet`
- `just precommit`


ops/production/compose.yml
```diff
@@ -23,7 +23,9 @@ services:
   postgres-analytics:
     image: postgres:16-alpine
     container_name: pensieve-postgres-analytics
-    restart: unless-stopped
+    # Retained for manual use only; excluded from normal boot-time compose up.
+    profiles: ["manual-postgres"]
+    restart: "no"
```
erskingardner: Nit: `"no"` is Compose's default restart policy, so this line could just be deleted. Removing `unless-stopped` has the same effect. Keep it if you want the intent spelled out. The comment above already explains it.

ops/production/compose.yml
```diff
@@ -204,5 +206,10 @@ networks:
   default:
     name: pensieve-deploy_default
     driver: bridge
+    # Match the existing network and host firewall rules after recreation.
```
erskingardner: The block comment just above still says the network is "reused instead of recreating it", and RUNBOOK step 5 checks "reused, NOT recreated". This new line implies it *was* recreated. Please reword one of them so they agree, e.g. "Pin IPAM to the live network's subnet/gateway so any recreation keeps the addresses the host firewall rules expect." Also worth confirming once on the host, beyond `config --quiet`: check that `docker network inspect pensieve-deploy_default` shows `172.18.0.0/16`/`172.18.0.1` and that `systemctl restart pensieve` doesn't try to recreate the network.

erskingardner (approved):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-23T14:21:56Z` |
| Commit reviewed | `4ec60df660489631da2db2f0632cabbbf7ec5d83` |

Small, focused change that brings the repo in line with what production is actually running: the API is back on ClickHouse and the Postgres refresh timer is disabled, so Postgres shouldn't start on boot. Pinning the subnet so that recreating the network doesn't break host firewall rules also makes sense. `config --quiet` passes, and the Rust CI is irrelevant to this file.

Non-blocking follow-ups:

- **The RUNBOOK and units now disagree with compose.** `ops/RUNBOOK.md` ("Shadow analytics Postgres" / "Recurring Slice A refresh") still describes a daily refresh. `ops/systemd/pensieve-analytics-refresh.service` relies on `Wants=pensieve.service` to have Postgres up. After this change, if someone re-enables the timer after a reboot, it fails every run, because nothing starts Postgres and `restart: "no"` keeps a manually started container down. Add a sentence to the RUNBOOK saying Postgres is manual-only and must be started first. `up -d postgres-analytics` still works because naming a service activates its profile. Alternatively, point readers at the retirement plan.
- See the inline notes on the redundant `restart` line and the network comment wording.

PR #45, opened 2026-09-23
Add durable negentropy job ledger foundation

erskingardner:
## Scope

First bounded increment of the separate-process negentropy architecture. Based on current master, independent of draft #42. No production changes and no runtime callers.

## Changes

- Dedicated SQLite WAL/FULL job ledger with canonical frozen relay/window identities and retained unresolved gaps.
- Serialized single-worker leases, random attempt capabilities, persisted backoff, stale/expired worker fencing, and atomic inclusive-window splits.
- Dense one-second windows remain blocked; there is deliberately no completion API until durable archive receipts exist.
- Runtime-adjustable row/byte admission budgets; no automatic pruning. The byte ceiling is admission control, not a hard filesystem quota.
- Scope, limitations, and remaining IPC/receipt/worker/rollout gates documented in docs/negentropy_job_ledger.md.

## Verification

- Ten new SQLite tests cover reopen, stale/forged leases, simultaneous claims, splitting, injected transaction rollback, budget rejection, and abrupt subprocess exit with committed and uncommitted work.
- Full just precommit passed immediately before commits 92f9085 and b9e89a3 (fmt, strict workspace clippy, workspace tests). Existing opt-in/ignored tests remain excluded by the standard suite.
- No production service, archive, database, or configuration touched.

## Explicit follow-ups

Authenticated bounded IPC and archive-confirmed receipt/completion accounting remain unimplemented. Scheduler fairness, 14-day planning, inventory, isolated worker, health alerts, Linux fault tests, and rollout are later slices. Do not treat this PR as a deployable replacement for the old scheduler.

## Current review status

One Claude review completed; the Grok review stalled and was terminated without findings. A single fix pass was pushed and passed full just precommit. No independent approval of the revised head; no merge or deployment. See the latest builder response for addressed and remaining findings.


crates/pensieve-ingest/src/sync/jobs.rs
```diff
@@ -0,0 +1,753 @@
...
+                        "INSERT INTO settings VALUES (1,?1,?2)",
+                        params![limits.max_jobs, limits.max_bytes as i64],
+                    )?;
+                    tx.pragma_update(None, "application_id", APPLICATION_ID)?;
+                    tx.pragma_update(None, "user_version", 1)?;
+                }
+                (APPLICATION_ID, 1) => {}
+                _ => return Err(LedgerError::Invalid("unsupported ledger identity/version")),
+            }
+            let saved: (u32, i64) = tx.query_row(
```
erskingardner: **Blocking: `Budget` is permanent by construction.** Three things combine here:
1. The limits are persisted, and reopening with different limits fails (`ledger limits differ`).
2. Rows are never deleted, and there's no completion state, no pruning and no VACUUM. Split parents and blocked jobs stay forever, and every new sweep adds rows.
3. The byte check reads file sizes (`file_bytes`, and the WAL's high-water mark), and those never shrink.

So once `count == max_jobs` or `bytes + reserve > max_bytes`, no operator action can clear it. An operator can't raise the limit, and the ledger can't shrink. The doc's "retaining existing jobs for inspection and recovery after capacity is restored" can't happen. With the defaults (100k rows) and the test's 900 s windows, a 14-day sweep is about 1,344 rows per relay, so roughly 75 relays fill the ledger in a single sweep.

Suggested fix, which also deletes code: drop the `settings` table and the limits-equality check, and treat `LedgerLimits` as runtime config the caller can raise. That removes about 12 lines here plus the `max_jobs: 1` reopen assertion in the tests.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-23T18:20:16Z` |
| Commit reviewed | `92f9085b41f4577922e2d57aec1d338b26e37d96` |
| Reasoning level | `15` |

> Posted as a comment because GitHub does not allow requesting changes on your own PR. The intended event is **Request changes**.

**Summary.** The transactional core holds up: leases are fenced by attempt and token, the expiry boundary is consistent (`expires_at > now` versus `<= now`), splits are gap-free and atomic, and the tests use real SQLite, including a subprocess-exit test. CI is green. One blocking design problem remains. The budget limits are frozen on disk, and nothing ever deletes or shrinks the ledger. On top of that, the byte check also gates `retry`, `expire` and `lease_next`. So once either ceiling is reached, the ledger stays stuck with no way out, which contradicts the doc's claim that jobs are retained "for inspection and recovery after capacity is restored". The fix makes the code smaller.

**Blocking**
- Persisted, immutable limits combined with append-only rows mean `Budget` is a terminal state (inline at L205).
- The byte-admission check gates state transitions that don't grow the ledger. When the budget is hit, the single global lease can never be recovered (inline at L318/L338).

**Non-blocking**
- The global `one_active_lease` index serializes every relay through one worker, and it is built into the v1 schema. At the test's 900 s window size, the throughput and row budget don't fit a 14-day, multi-relay plan (inline at L192).
- Several statements are hand-written on single long lines that rustfmt skips, so `just fmt` never formats them (inline at L248).
- The `attempt == i64::MAX` guard and its test cover a situation that cannot happen (inline at L286).
- Building the split children with a `Vec` and two `remove(0)` calls is clunky (inline at L370).

**Premise.** Staging the work is reasonable, and #46 already builds on it. But merging a versioned on-disk schema (`application_id`, v1) with no callers locks in decisions like the persisted limits and the global single lease before any real caller has tested them. #46 already needs a v1 to v2 migration. Consider landing #45 and #46 together, or keeping the schema unversioned until the first runtime caller exists.

Validation: I relied on the existing CI (fmt, clippy, doctests, 3 test shards all pass). I ran nothing locally except `rustfmt` on a /tmp copy of `jobs.rs`, which confirmed that rustfmt leaves the long SQL statements unchanged.

crates/pensieve-ingest/src/sync/jobs.rs
```diff
@@ -0,0 +1,753 @@
...
+        now: i64,
+        delay_secs: u32,
+        reason: RetryReason,
+    ) -> Result<(), LedgerError> {
+        let due = retry_due(now, delay_secs)?;
+        let tx = self
+            .db
+            .transaction_with_behavior(TransactionBehavior::Immediate)?;
+        verify_lease(&tx, lease, now)?;
+        check_budget(&tx, &self.path, self.limits, 0)?;
```
erskingardner: **Blocking: the byte admission check shouldn't gate non-growing transitions.** `retry` (here), `expire` (L338) and `lease_next` (L285) are UPDATEs on an existing row. They don't admit new work. But `check_budget` runs in each of them and again inside `commit()`. So once the byte ceiling is crossed, a live lease can't be retried, and an expired lease can't be recovered by `expire`. Because `one_active_lease` allows only one lease globally, that stuck row blocks `lease_next` for every relay.

Admission control belongs on `enqueue` and the non-dense `split`, which are the operations that add rows. Suggest removing the budget checks from `retry`, `expire` and `lease_next`. Since `commit()` checks too, give those three a plain `tx.commit()`. `row_and_byte_limits_preserve_existing_obligations` should then assert the opposite: `retry` succeeds when over budget, and `enqueue` is rejected.

crates/pensieve-ingest/src/sync/jobs.rs
```diff
@@ -0,0 +1,753 @@
...
+                        CREATE TABLE jobs (
+                            id INTEGER PRIMARY KEY, sweep TEXT NOT NULL, relay TEXT NOT NULL,
+                            since INTEGER NOT NULL CHECK(since>=0), until INTEGER NOT NULL CHECK(until>=since),
+                            parent INTEGER REFERENCES jobs(id), state TEXT NOT NULL CHECK(state IN ('queued','leased','retry_wait','blocked','split')),
+                            attempt INTEGER NOT NULL DEFAULT 0 CHECK(attempt>=0), token BLOB, expires_at INTEGER,
+                            next_eligible INTEGER NOT NULL DEFAULT 0, reason TEXT,
+                            UNIQUE(sweep,relay,since,until),
+                            CHECK((state='leased' AND token IS NOT NULL AND typeof(token)='blob' AND length(token)=32 AND expires_at IS NOT NULL) OR (state!='leased' AND token IS NULL AND expires_at IS NULL))
+                        );
+                        CREATE UNIQUE INDEX one_active_lease ON jobs(state) WHERE state='leased';
```
erskingardner: Non-blocking design question: this partial unique index allows exactly one lease **across all relays and sweeps**, and it's part of the v1 schema. Some rough numbers: at the test's 900 s windows, a 14-day lookback is 1,344 windows per relay. At about 5 s per negentropy exchange, 100 relays is around 134k serial leases, roughly 7.8 days, before counting retries and splits. That's longer than the lookback it's trying to cover, and it's also over the default `max_jobs`.

If single-worker is a deliberate first-rollout choice, it could be enforced in `lease_next` (which already checks `active`) rather than in the schema. Then moving to per-relay concurrency later wouldn't need a migration. If it's meant to be permanent, the doc should include this sizing math.

crates/pensieve-ingest/src/sync/jobs.rs
```diff
@@ -0,0 +1,753 @@
...
+        {
+            return Err(LedgerError::Invalid("invalid window or identity"));
+        }
+        let relay = crate::relay::normalize_relay_url(relay)
+            .ok()
+            .ok_or(LedgerError::Invalid("invalid relay"))?;
+        let tx = self
+            .db
+            .transaction_with_behavior(TransactionBehavior::Immediate)?;
+        if let Some(id) = tx.query_row("SELECT id FROM jobs WHERE sweep=?1 AND relay=?2 AND since=?3 AND until=?4 AND parent IS NULL", params![sweep,relay,since,until], |r| r.get::<_,i64>(0)).optional()? {
```
erskingardner: Non-blocking readability: this line and L251, L283, L372, L394, L402 and L410-418 exceed rustfmt's width because of the inline SQL string. rustfmt skips those statements without any warning, so they never get formatted. That's why they read like minified code (`let id=tx.query_row(...)`, `"queued"=>JobState::Queued,"leased"=>...` on one line). I confirmed this: running `rustfmt --edition 2024` on this file produces no diff.

Pulling the SQL into `const` raw strings, or putting the literal on its own line, lets rustfmt handle these, and the project's default-`cargo fmt` convention then covers the whole file.

crates/pensieve-ingest/src/sync/jobs.rs
```diff
@@ -0,0 +1,753 @@
...
+            [],
+            |r| r.get(0),
+        )?;
+        if active {
+            return Ok(None);
+        }
+        let id=tx.query_row("SELECT id FROM jobs WHERE state IN ('queued','retry_wait') AND next_eligible<=?1 ORDER BY id LIMIT 1",[now],|r|r.get::<_,i64>(0)).optional()?;
+        let Some(id) = id else { return Ok(None) };
+        check_budget(&tx, &self.path, self.limits, 0)?;
+        if read_job(&tx, id)?.attempt == i64::MAX {
```
erskingardner: Simplification: this guards against a job being leased 9.2e18 times, which can't happen. And if it somehow did happen, it would be worse than no guard: this is the oldest eligible row, so `lease_next` would return `Err` on every call and starve the whole queue. Suggest deleting this check along with the second half of `invalid_capability_and_attempt_overflow_fail_closed`.

crates/pensieve-ingest/src/sync/jobs.rs
```diff
@@ -0,0 +1,753 @@
...
+                } else {
+                    "resource_split"
+                }
+            ],
+        )?;
+        let outcome = if dense {
+            SplitOutcome::Blocked
+        } else {
+            let midpoint = parent.since + (parent.until - parent.since) / 2;
+            let mut children = Vec::with_capacity(2);
```
erskingardner: Nit: the `Vec::with_capacity(2)` plus a loop plus two `children.remove(0)` calls is a roundabout way to build two values. A small `insert_child(since, until)` closure called twice, giving `Box::new([insert_child(parent.since, midpoint)?, insert_child(midpoint + 1, parent.until)?])`, reads more directly and drops the temporary `Vec`.

erskingardner:
Builder response to this round at 92f9085: fixed the recovery blockers in b9e89a3. Limits are runtime policy and can be raised after capacity review; retry and expiry bypass admission ceilings and preserve existing obligations. Tests prove recovery while over budget and subsequent leasing after reopening with restored capacity. New leases still respect the byte ceiling intentionally: releasing old ownership must remain possible, but starting more work under disk pressure is not recovery. Full just precommit passed immediately before commit.

The single-worker constraint is an explicitly approved initial canary policy, not a throughput claim for an unrestricted relay catalog. Keeping its database invariant and the fail-closed overflow guard; cosmetic suggestions deferred. This does not enable the runtime.

Review execution: Claude completed and posted its intended request-changes verdict as COMMENT because GitHub prevents author self-request-changes. The Grok reviewer stalled with an empty log and no posted review, was terminated/reaped (exit 143), and was not retried. This is not two completed independent reviews or approval of the revised head. PR remains unmerged.

crates/pensieve-ingest/src/sync/jobs.rs
```diff
@@ -0,0 +1,745 @@
...
+    #[error("ledger admission budget exhausted")]
+    Budget,
+}
+
+/// Independent limits on job rows and SQLite/WAL admission footprint.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct LedgerLimits {
+    /// Includes split parents; no automatic history deletion.
+    pub max_jobs: u32,
+    /// Checked before mutations and before commit, including WAL, shared memory,
```
erskingardner: Nit: this doc is now inaccurate. `retry` and `expire` skip the byte check on purpose, and the doc should say so. Suggest: `/// Checked before admitting work (enqueue, lease, split) and before commit; retry/expiry bypass it so a lease can always be released.`

crates/pensieve-ingest/src/sync/jobs.rs
```diff
@@ -0,0 +1,745 @@
...
+                Err(LedgerError::Invalid(_))
+            ));
+        }
+        for (now, delay) in [(-1, 60), (0, 59), (0, 3601), (i64::MAX, 60)] {
+            assert!(matches!(
+                db.expire(now, delay),
+                Err(LedgerError::Invalid(_))
+            ));
+        }
+        assert!(
```
erskingardner: Nit: this assertion flipped from rejecting different limits to accepting them. After that change, nothing exercises the `max_jobs == 0` / `max_bytes < 4 * WRITE_RESERVE` guard at L159-164, but the test name still says `..._limits_...`. Suggest turning this into the invalid-limits case, e.g. `LedgerLimits { max_jobs: 0, ..Default::default() }` → `Err(Invalid)`. Reopening with changed limits is already covered at L618 and L625.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-23T19:28:39Z` |
| Commit reviewed | `b9e89a3300fdb3755f902a25382c53b77dfe6771` |

> Posted as a comment because GitHub does not allow approving your own PR. The intended event is **Approve**.

**Summary.** Both blocking findings from the previous round are fixed, and the fix removed code: net -2 lines in `jobs.rs`.

- The `settings` table and the "ledger limits differ" check are gone. Limits are now runtime policy, so an operator can reopen with larger limits and get out of `Budget`.
- `retry` and `expire` no longer call `check_budget`, and they commit directly. The single global lease can always be released. The rewritten `row_and_byte_limits_preserve_existing_obligations` test covers the whole path: retry while over the ceiling, reopen, expire while over the ceiling, `lease_next` rejected, then a reopen with restored limits that leases again.
- `lease_next` still checks the byte ceiling. I agree with that choice: starting new work under disk pressure isn't recovery.

Changing the v1 schema in place without a version bump is fine because nothing has shipped. #46 already merges these fixes (1b2d0b2).

**Non-blocking (inline)**
- The `LedgerLimits::max_bytes` doc still says the limit is checked "before mutations". That's no longer true for `retry` and `expire`.
- No test covers the invalid-limits guard in `open` any more, and the test name still says it rejects limits.

The earlier non-blocking items are acknowledged and deferred by the author, and I'm not raising them again: the single-lease index as canary policy, rustfmt-skipped SQL lines, the `i64::MAX` guard, and `remove(0)`.

Validation: I relied on the existing CI at this head. fmt, clippy, doctests and all 3 test shards pass. I ran nothing locally.

erskingardner:
Fresh review round on b9e89a3: Claude recommends merge, with only nonblocking documentation/test nits. GitHub recorded COMMENTED rather than APPROVED because the review uses the author account. The Grok seat did not post a #45 review; its process was stopped/reaped with exit 143 after about 5.5 minutes and an empty log. No blocking change is needed on #45 from this round. Leaving this PR unmerged as requested while the archive consumer is developed above #46.

PR #46, opened 2026-09-23
Add bounded negentropy upload and received receipts

erskingardner:
## Scope
Bounded isolated-worker upload library on master, following merged #45. No listener, worker, deployment, SDK changes or production migration.

Received receipts commit before handoff; framing, signature/window checks, credit and attempt budgets fail closed. ProtocolDone checks persisted count, digest and sequence as well as session ordering. Reconnection cannot summarize another session’s receipts. Archive completion is implemented by companion #48; land the two together, with runtime still disabled.

## Review and verification
Claude Opus 5.5 and Grok 4.7 both reviewed 6064c03. Their current-round findings were addressed: the branch now rests on the squash-merged master foundation, and a regression covers reconnected ProtocolDone. Full just precommit passed immediately before ee873d8. Fresh CI is required on this revised head; do not interpret prior-head reviews as approval of unseen changes.

Both reviewer processes exited successfully. Earlier reports that quiet Grok output meant a stalled process were incorrect; this round used streaming output.

See docs/negentropy_upload_protocol.md for the wire contract and remaining integration gates.


crates/pensieve-ingest/src/sync/jobs.rs
```diff
@@ -210,6 +214,26 @@ impl JobLedger {
             if saved != (limits.max_jobs, limits.max_bytes as i64) {
                 return Err(LedgerError::Invalid("ledger limits differ"));
             }
+            if version < 2 {
```
erskingardner: #45 hasn't merged and no v1 ledger has been deployed (the PR says "no production migration"), so nothing uses this migration. Add `attempts` (and `receipts`, if it stays) to the `(0, 0)` create batch. That lets you delete this branch, the `1 | 2` match arm, the double `check_budget` and `v1_migration_preserves_jobs_and_active_capability`. Add a real migration once a ledger exists somewhere.

crates/pensieve-ingest/src/sync/jobs.rs
```diff
@@ -210,6 +214,26 @@ impl JobLedger {
...
+                check_budget(&tx, path, limits, 0)?;
+                tx.execute_batch(
+                    "CREATE TABLE attempts (
+                        job INTEGER NOT NULL REFERENCES jobs(id), attempt INTEGER NOT NULL,
+                        received INTEGER NOT NULL, bytes INTEGER NOT NULL, digest BLOB NOT NULL,
+                        protocol_done INTEGER NOT NULL DEFAULT 0,
+                        PRIMARY KEY(job,attempt)
+                    );
+                    CREATE TABLE receipts (
```
erskingardner: Nothing in this PR reads this table's columns. `event_id` and `created_at` are never queried. The only reads are the idempotent-retry lookup below (see the next comment) and the global `count(*)` for its own cap. `attempts` already persists count, bytes and digest. I'd add `receipts` together with the archive receipt-satisfaction code that consumes it, so its schema is shaped by a real reader. That would also remove `MAX_RECEIPTS` and the 100k-row cap, which as documented would stop the whole ledger after about two full-size attempts.

crates/pensieve-ingest/src/sync/jobs.rs
```diff
@@ -382,6 +406,155 @@ impl JobLedger {
...
+        {
+            return Err(LedgerError::Invalid("invalid receipt metadata"));
+        }
+        let progress = attempt_progress(&tx, lease.job.id, lease.job.attempt)?;
+        if progress.protocol_done {
+            return Err(LedgerError::Invalid("protocol already ended"));
+        }
+        // Exact retries cover uncertain commit acknowledgement. The network
+        // session still rejects duplicate frames; it cannot release credit twice.
+        let existing = tx
```
erskingardner: No caller exercises this exact-retry path. `UploadSession::receive` poisons the session on any error, and nothing retries `register_received`. With SQLite, a failed commit returns `Err`, so there is no uncertain acknowledgement in-process. The branch also opens the reconnected-session replay hole that `receive` has to close with its extra `attempt_progress` guard (ipc.rs:343). Without it, the `sequence != progress.received + 1` check just below already rejects a replay. Suggest deleting both, and the matching early `Ok(())` in `record_protocol_done` (line 507).

crates/pensieve-ingest/src/sync/jobs.rs
```diff
@@ -382,6 +406,155 @@ impl JobLedger {
...
+                Err(LedgerError::Invalid("conflicting receipt retry"))
+            };
+        }
+        if received.sequence != progress.received + 1
+            || received.sequence > super::ipc::MAX_EVENTS
+            || progress.bytes.saturating_add(received.frame_bytes) > super::ipc::MAX_ATTEMPT_BYTES
+        {
+            return Err(LedgerError::Invalid("receipt sequence or attempt limit"));
+        }
+        let count: i64 = tx.query_row("SELECT count(*) FROM receipts", [], |r| r.get(0))?;
```
erskingardner: This is a full `count(*)` scan over up to 100k rows on every event insert. Together with two `check_budget` calls (each a jobs `count(*)` plus three `stat`s) and a FULL fsync, each received event costs a lot. If the table stays, keep a counter in `settings` or derive the cap from `attempts`. Preferably, drop the table for now (see above).

crates/pensieve-ingest/src/sync/ipc.rs
```diff
@@ -0,0 +1,810 @@
...
+            || header.token != self.lease.token()
+            || header.sequence != self.count + 1
+            || now < 0
+            || now >= self.lease.expires_at()
+        {
+            return Err(ProtocolError::State);
+        }
+        let frame_bytes = frame.wire_bytes();
+        let frame_digest = frame.digest();
+        ledger.verify_active(&self.lease, now)?;
```
erskingardner: This check is redundant here. `register_received` and `record_protocol_done` both run `verify_lease` inside their own transactions, and for `AttemptFailed` the caller's `retry` checks the lease anyway. It adds an extra IMMEDIATE (write-lock) transaction per event only to read. Keeping it in `accepted` makes sense, because that is the only persisted check on the ACK path.

crates/pensieve-ingest/src/sync/ipc.rs
```diff
@@ -0,0 +1,810 @@
...
+                }
+                let timestamp = event.created_at.as_secs();
+                if timestamp < self.lease.job().since as u64
+                    || timestamp > self.lease.job().until as u64
+                    || event.verify().is_err()
+                {
+                    return Err(ProtocolError::Event);
+                }
+                // A new session cannot replay an already partially received attempt.
+                let prior = ledger.attempt_progress(header.job, header.attempt)?;
```
erskingardner: This guard exists only because `register_received` accepts exact retries of an existing sequence (jobs.rs:446). Remove that branch and the ledger's own `sequence == received + 1` and `protocol_done` checks reject a reconnected replay. This block and one DB read per event can then go.

crates/pensieve-ingest/src/sync/ipc.rs
```diff
@@ -0,0 +1,810 @@
...
+            || event.header.attempt != self.lease.job().attempt
+        {
+            self.closed = true;
+            return Err(ProtocolError::State);
+        }
+        if let Err(error) = ledger.verify_active(&self.lease, now) {
+            self.closed = true;
+            return Err(error.into());
+        }
+        let entry = self.outstanding.first_key_value();
```
erskingardner: Nit: ACKs are strictly FIFO, so a `VecDeque<(u64, u64, [u8; 32])>` fits better than a `BTreeMap`. Checking `front()` and then calling `pop_front()` also removes the unreachable `remove(..).ok_or(State)?` below, which is the one error path here that doesn't set `closed`.

crates/pensieve-ingest/src/sync/ipc.rs
```diff
@@ -0,0 +1,810 @@
...
+        reader.read_exact(&mut prefix)?;
+        let length = u32::from_be_bytes(prefix) as usize;
+        if length == 0 || length > MAX_FRAME_BYTES {
+            return Err(ProtocolError::Limit);
+        }
+        let mut payload = vec![0; length];
+        reader.read_exact(&mut payload)?;
+        let message: Message =
+            serde_json::from_slice(&payload).map_err(|_| ProtocolError::Malformed)?;
+        if message.header().version != VERSION {
```
erskingardner: Nit: `receive` checks the version again (line 314). One check is enough. I'd keep the one in `receive`, since `Frame` values can only come from `read`.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-23T18:19:10Z` |
| Commit reviewed | `867b5d31ecf9f9b95b32d724a2b37bd9650b62b1` |
| Reasoning level | `reasoning_effort=15` |

## Summary

The codec parts are solid: the length is checked before allocation, encoding is capped, unknown fields are rejected, the sequence must be strict, and every error poisons the session. The tests are thorough, and CI passes on this SHA. My concern is how much speculative durable state the PR adds on top of #45, which also has no callers.

**Premise.** This is a second unwired layer (about 1.2k lines), and it fixes a schema and wire format before the consumer, archive receipt satisfaction, exists. As written it also cannot run continuously. The global cap is 100,000 retained receipts with no pruning, and one attempt can use 50,000. Two full-size attempts would leave every relay's ledger returning `Budget`. The doc says so. That is a strong sign the receipt design should land with the code that reads and compacts it, not before.

**Could be much smaller (net deletions):**
1. **Drop the v1 to v2 migration.** #45 is unmerged and has never been deployed ("no production migration"), so no v1 ledger exists to migrate. Put the new tables in the base schema. That removes the `version < 2` branch, the `1 | 2` arm and `v1_migration_preserves_jobs_and_active_capability`.
2. **The `receipts` table is write-only.** Nothing reads `event_id`, `created_at`, `frame_bytes` or `frame_digest`. Its only readers are the idempotent-retry lookup (item 3) and a `count(*)` for its own cap. The `attempts` row already holds count, bytes and digest. Add receipts when archive satisfaction needs them, and `MAX_RECEIPTS` and the per-insert scan go with them.
3. **The exact-retry idempotency is dead code that forces a second guard.** Every caller poisons the session on any error, and none retries `register_received`. The branch is also why `receive` needs its own `attempt_progress` replay check. Without it, the ledger's `sequence != received + 1` check already rejects a reconnected replay. Remove both. The same applies to the early `Ok(())` for a second `protocol_done`.
4. **`verify_active` in `receive` is redundant.** `register_received` and `record_protocol_done` each check the lease again, and the caller's `retry` checks it for `AttemptFailed`.

**Hot-path cost (if the design stays).** Each event currently opens two IMMEDIATE transactions, does two or three progress reads, scans up to 100k receipt rows with `count(*)`, runs `check_budget` twice (a jobs `count(*)` plus three `stat` calls each time), and does a FULL fsync. A 50k-event window pays that 50k times. Items 2 to 4 remove most of it.

Details are in the inline comments. No correctness bugs found; everything above is about scope and simplicity.

erskingardner:
Builder response to this round at 867b5d3: addressed the actionable simplification/performance findings in 1b2d0b2, incorporating the #45 recovery fix. Removed the speculative v1 migration, exact-retry acceptance and compensating progress read, and redundant event-path lease transaction. Incompatible prototype schemas fail closed without mutation. Receipt count admission now uses a transactionally maintained counter instead of scanning all receipt rows per event. ACKs and failure messages still validate persisted ownership. Replay, rollback, reopen, admission limits and the complete just precommit gate pass.

Remaining concern, not claimed resolved: receipt consumption/terminal compaction and throughput validation must land before runtime use. I retained per-event identity receipts because the approved archive-first contract requires accounting for each received ID; dropping them in favor of a digest summary would defer that safety boundary, not establish archive durability. Per-event FULL commit cost remains unbenchmarked. Recommend holding this PR unmerged until its archive consumer/compaction integration can be evaluated, rather than claiming the current write-only foundation is operationally complete.

Review execution: Claude posted serious scope concerns; Grok stalled without output or a GitHub review and was terminated/reaped (exit 143). No second review round was launched. The revised head is not independently approved. Production remains untouched.

crates/pensieve-ingest/src/sync/ipc.rs
```diff
@@ -0,0 +1,804 @@
...
+            || now < 0
+            || now >= self.lease.expires_at()
+        {
+            return Err(ProtocolError::State);
+        }
+        let frame_bytes = frame.wire_bytes();
+        let frame_digest = frame.digest();
+        match frame.message {
+            Message::Event { header, event } => {
+                if self.count >= MAX_EVENTS
```
erskingardner: The ledger now enforces these limits authoritatively inside the same transaction. `register_received` checks `sequence > MAX_EVENTS` and `progress.bytes + frame_bytes > MAX_ATTEMPT_BYTES`, and `record_protocol_done` checks the count and digest against persisted progress. Every ledger error already poisons the session, so the in-memory `bytes` and `digest` fields are shadow copies of state the ledger owns. So are the `MAX_EVENTS`/`MAX_ATTEMPT_BYTES` checks here and the `count != self.count || digest != self.digest` check at line 366. Keep `count` for header sequencing and the in-flight credit checks, and drop the rest. The alternative is to drop the ledger's duplicates, including the window and frame-size re-checks at jobs.rs:421-428. Either way, one layer should own each invariant.

crates/pensieve-ingest/src/sync/jobs.rs
```diff
@@ -371,6 +394,147 @@ impl JobLedger {
...
+        }
+        let count: i64 = tx.query_row(
+            "SELECT retained FROM receipt_totals WHERE singleton=1",
+            [],
+            |r| r.get(0),
+        )?;
+        if count >= MAX_RECEIPTS {
+            return Err(LedgerError::Budget);
+        }
+        check_budget(&tx, &self.path, self.limits, 0)?;
```
erskingardner: `commit()` already runs `check_budget` after the writes, and that is the stricter check because it sees the grown page count. This pre-check with `additional = 0` cannot reject anything the post-write check would accept. On the per-event path it doubles the jobs `count(*)` and the three `stat` calls. The same applies at line 490 in `record_protocol_done`. #45 uses the same pattern, but this is the hot path, so I'd drop both pre-checks here.

crates/pensieve-ingest/src/sync/jobs.rs
```diff
@@ -371,6 +394,147 @@ impl JobLedger {
...
     }
+
+    /// Bounded persisted accounting for one attempt. Receipt registration is not
+    /// evidence of archive durability or novelty. No payloads are stored here.
+    pub fn attempt_progress(&self, job: i64, attempt: i64) -> Result<AttemptProgress, LedgerError> {
+        attempt_progress(&self.db, job, attempt)
+    }
+
+    pub(super) fn verify_active(&mut self, lease: &Lease, now: i64) -> Result<(), LedgerError> {
+        let tx = self
```
erskingardner: `verify_lease` is a single `SELECT EXISTS(...)`, and one statement is already atomic. Taking an IMMEDIATE transaction acquires the write lock on every ACK only to roll it back. A deferred `self.db.transaction()` would be enough, or `verify_lease` could take `&Connection`, since `Transaction` derefs to it.

crates/pensieve-ingest/src/sync/jobs.rs
```diff
@@ -739,7 +904,168 @@ mod tests {
...
+            .execute(
+                "UPDATE attempts SET received=1,bytes=?1 WHERE job=?2",
+                params![super::super::ipc::MAX_ATTEMPT_BYTES, lease.job.id],
+            )
+            .unwrap();
+        assert!(db.register_received(&lease, &receipt(2), 12).is_err());
+        db.db
+            .execute("UPDATE attempts SET bytes=500 WHERE job=?1", [lease.job.id])
+            .unwrap();
+        // Seed retained historical rows to exercise the global admission bound.
```
erskingardner: Since the cap now reads `receipt_totals`, these approximately 100k seeded `receipts` rows are never consulted. They only slow the test down, and the comment above no longer describes what is being exercised. The `UPDATE receipt_totals SET retained=?1` below is sufficient on its own, so delete this insert.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-23T19:29:29Z` |
| Commit reviewed | `1b2d0b2e8d20c4602af02fdbdcd4e3b9a7ee59a8` |
| Reasoning level | `reasoning_effort=15` |

## Summary

This is a re-review of `1b2d0b2`, which comes after my review of `867b5d3`. The fix pass is a real improvement and is also a net deletion:
- The speculative v1 migration is gone.
- Exact-retry acceptance, along with the replay guard it made necessary, is gone.
- The redundant `verify_active` on the event path is gone.
- The per-insert `count(*)` scan is now a transactional counter.

I found no correctness bugs, and CI passes on this SHA.

**Premise (still the main concern).** The remaining objection is the one the builder also raised: the `receipts` table is still write-only. Nothing reads `event_id`, `created_at`, `frame_bytes` or `frame_digest`. Its schema is fixed before its consumer (archive satisfaction and compaction) exists. The 100k global cap is also reached after two full-size attempts, with no pruning path. I understand the argument about per-ID accounting under the archive-first contract. But committing a schema before the code that reads it means that schema will likely need changing once the consumer exists. I agree with the builder's recommendation to **hold this PR unmerged** until the archive consumer and compaction can land with it, or can at least be evaluated against it. I would not merge this layer on its own.

**Smaller remaining simplifications (inline):**
1. Attempt limits and the summary are enforced twice: once in the in-memory `UploadSession` and again in the persisted ledger. Keep them in one layer.
2. `register_received` and `record_protocol_done` each run `check_budget` before the writes, and `commit()` runs it again. On the per-event hot path, that doubles a jobs `count(*)` plus three `stat` calls.
3. `verify_active` takes an IMMEDIATE write lock for a single read on every ACK.
4. The global-ceiling test still inserts about 100k receipt rows that the counter-based cap never reads.

My earlier nits (a FIFO `VecDeque` instead of the `BTreeMap`, and the duplicate version check) still apply to this head.

crates/pensieve-ingest/src/sync/jobs.rs
```diff
@@ -371,6 +394,147 @@ impl JobLedger {
     pub fn get(&self, id: i64) -> Result<Job, LedgerError> {
         read_job(&self.db, id)
     }
+
+    /// Bounded persisted accounting for one attempt. Receipt registration is not
+    /// evidence of archive durability or novelty. No payloads are stored here.
+    pub fn attempt_progress(&self, job: i64, attempt: i64) -> Result<AttemptProgress, LedgerError> {
```
erskingardner: The crash-before-admission case this table is meant to cover is still not recoverable from here. This is the only read API, and it returns the aggregate on `attempts` (count, bytes, digest). It never returns the stored event IDs, so nothing in this PR can reconcile, satisfy, or delete a committed receipt.

`MAX_RECEIPTS` (100_000) is a hard-coded constant, separate from `LedgerLimits`. Reopening with a higher job or byte ceiling does not raise it. A saturated attempt is retained across expiry and retry, and `ProtocolDone` does not free those rows, so a second full attempt makes every later `register_received` return `Budget` for good.

The sequence, byte, and digest checks already live on `attempts`. I would keep that summary with the codec and add `receipts` only in the change that lists those IDs and compacts them.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-23T19:32:46Z` |
| Commit reviewed | `1b2d0b2e8d20c4602af02fdbdcd4e3b9a7ee59a8` |
| Model | `Grok 4.7` |

## Summary

The fix pass holds up. Incompatible v1 ledgers fail closed with no migration, replays are rejected, admission uses a maintained counter, and the extra lease transaction is gone from the event path. CI is green on this SHA. The codec, sequence checks, and rollback tests look correct.

This still should not merge. Per-event receipts have no reader, `ProtocolDone` leaves the job leased, and the 100,000-row ceiling is a hard-coded constant. Two full attempts stop every later registration, and reopening with a higher job or byte limit does not move that ceiling. The bounded codec and the `attempts` summary are enough for this increment; add `receipts` in the change that can list, satisfy, and compact them.

The open nits on the duplicate version check and the FIFO `BTreeMap` are non-blocking. The `jobs` module docs still say receipt accounting belongs to a later increment.

erskingardner:
Addressed this round at 6064c03: persisted ledger now solely owns attempt byte/count/digest enforcement; ACK lease verification is a single read rather than an IMMEDIATE write transaction; upload mutations use one post-write budget check; the cap test seeds only its authoritative counter. Full just precommit passed immediately before commit. Keeping the per-ID receipts because the approved archive-first contract requires recovering obligations after registration; the next PR stacked here will provide the real archive consumer, completion, and bounded compaction. Hold this PR unmerged until that consumer is reviewed with it. Both Claude and Grok reviews landed on 1b2d0b2; Grok process required termination (143) with an empty log after posting. No production changes.

crates/pensieve-ingest/src/sync/ipc.rs
```diff
@@ -0,0 +1,816 @@
...
+                ledger.verify_active(&self.lease, now)?;
+                Ok(UploadAction::Failed(reason))
+            }
+        }
+    }
+
+    /// Release credit only after the caller establishes admission ownership.
+    /// Receipt registration alone cannot establish that future runtime guarantee.
+    /// ACKs are ordered, single-use, and do not imply archival or completion.
+    pub fn accepted(
```
erskingardner: #48, which is stacked on this PR, makes this private behind `admit_and_accept`. It does that so that no caller can release credit without archive admission. If #46 merges on its own, master ships the public ACK-without-admission API that #48 deliberately removes. That is one more reason to land the two together.

crates/pensieve-ingest/src/sync/jobs.rs
```diff
@@ -0,0 +1,1059 @@
...
+                        PRIMARY KEY(job,attempt,sequence),
+                        FOREIGN KEY(job,attempt) REFERENCES attempts(job,attempt)
+                    );
+                    CREATE TABLE receipt_totals (
+                        singleton INTEGER PRIMARY KEY CHECK(singleton=1),
+                        retained INTEGER NOT NULL CHECK(retained>=0)
+                    );
+                    INSERT INTO receipt_totals VALUES(1,0);")?;
+                    tx.pragma_update(None, "application_id", APPLICATION_ID)?;
+                    tx.pragma_update(None, "user_version", 2)?;
```
erskingardner: #48 bumps this to 3 and rejects v2 as "the previous undeployed receipt schema". That means v2 would exist on master only between the two merges and would never reach a deployment. If they land back to back or folded together, #48 can keep version 2 and edit the tables in place. Its v2-rejection test and the version bump can then be deleted.

crates/pensieve-ingest/src/sync/jobs.rs
```diff
@@ -0,0 +1,1059 @@
...
+                    check_budget(&tx, path, limits, 0)?;
```
erskingardner: This budget check on a freshly created ledger is new in this PR (#45 had none here), and no test covers it. It looks like a leftover from the removed v1-to-v2 migration branch. The schema is only a few pages, and `open` already rejects `max_bytes < 4 * WRITE_RESERVE`. I would delete it, or add a test if it is meant to guard a real case.

crates/pensieve-ingest/src/sync/jobs.rs
```diff
@@ -0,0 +1,1059 @@
...
+        &mut self,
+        lease: &Lease,
+        received: &super::ipc::ReceiptRecord,
+        now: i64,
+    ) -> Result<(), LedgerError> {
+        let tx = self
+            .db
+            .transaction_with_behavior(TransactionBehavior::Immediate)?;
+        verify_lease(&tx, lease, now)?;
+        if received.created_at < lease.job.since as u64
```
erskingardner: `ReceiptRecord` is `pub(super)` and only `UploadSession::receive` builds it. That happens right after the same `since`/`until` check (ipc.rs:329), with a `frame_bytes` that `Frame::read` has already bounded. The last round moved attempt limits into the ledger so they would be enforced in one layer. The same reasoning applies here: keep the window check in one place and drop the other.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-23T20:07:19Z` |
| Commit reviewed | `6064c03ab9e9f2d15293a2061950fa0c0916031e` |
| Reasoning level | `reasoning_effort=15` |

## Summary

This is a re-review at `6064c03`. The fix pass after `1b2d0b2` landed cleanly:
- The in-memory attempt limits are gone.
- The ACK lease check is now a plain read.
- The budget check runs once.
- The cap test seeds only the counter.

I found no new correctness bugs in the delta over #45, and CI is green on this SHA.

**Blocking: the PR no longer merges.** #45 was squash-merged as `ce7c308`, whose tree is identical to `b9e89a3`. This branch still carries #45's original commits. As a result, GitHub reports `CONFLICTING`/`DIRTY`, and the diff shown is +2052 lines, about 1.2k of which are #45 again. `git rebase --onto origin/master b9e89a3` should apply cleanly because the trees match. After the rebase, update the stale PR body ("Current head 1b2d0b2", "Depends on #45"), the "Stacked on ... PR #45" line in `docs/negentropy_upload_protocol.md`, and #48's base.

**Stack: land this with #48, not alone.** My earlier main concern was that the receipts table had no reader. #48 answers that with `reconcile_archived` and compaction. But #48 also reverses parts of this PR:
- It makes `accepted` private behind `admit_and_accept`.
- It replaces `MAX_RECEIPTS` with an adjustable capacity.
- It bumps `user_version` from 2 to 3 and rejects v2 as "the previous undeployed receipt schema".

Merged alone, #46 would put two things on master: a schema version that will never be deployed, and the public ACK-without-admission API that the next PR removes on purpose. I agree with the builder's own hold. Rebase, then either merge #46 and #48 back to back or fold them together. In both cases #48 can keep version 2 instead of bumping it.

**Non-blocking (inline):**
- The event-window check is duplicated.
- `check_budget` on ledger creation is untested.
- My earlier nits still apply: use a FIFO `VecDeque` (ipc.rs:394), and drop the duplicate version check (ipc.rs:147).
- `docs/negentropy_upload_protocol.md:52` mentions "merge approval". That is process state and doesn't belong in a design doc.

crates/pensieve-ingest/src/sync/ipc.rs
```diff
@@ -0,0 +1,816 @@
...
+                    event,
+                    header,
+                    digest: frame_digest,
+                }))
+            }
+            Message::ProtocolDone { count, digest, .. } => {
+                if !self.outstanding.is_empty() {
+                    return Err(ProtocolError::State);
+                }
+                ledger.record_protocol_done(&self.lease, count, digest, now)?;
```
erskingardner: A second `UploadSession` starts `count` at 0, so the check above accepts `header.sequence == 1`. This arm then forwards the worker's `count` and `digest` and only requires the in-memory ACK map to be empty. `record_protocol_done` never sees the sequence.

After receipts have already committed, that second session can send `ProtocolDone { sequence: 1, count: persisted.received, digest: persisted.digest }` and set `protocol_done` without registering those frames or draining their ACKs. A worker that continues at sequence `received + 1` is rejected. The case that succeeds is a sender that restarts numbering at 1 and quotes the stored summary. `duplicate_frame_and_reconnected_partial_attempt_fail_closed` only replays an event frame.

Pass `header.sequence` into `record_protocol_done` and reject unless `sequence == progress.received + 1`, the same rule as event frames. On one session `self.count == progress.received`, so a real `ProtocolDone` still uses sequence `count + 1`. A fresh session can only present sequence 1, which matches only an empty attempt. Extend the reconnect test with this `ProtocolDone` and assert `protocol_done` stays false.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-23T20:10:31Z` |
| Commit reviewed | `6064c03ab9e9f2d15293a2061950fa0c0916031e` |
| Model | `Grok 4.7` |

## Summary

The `6064c03` pass lands the last round's simplifications. The ledger alone enforces attempt count, bytes, and digest. ACK lease checks are a single read. Upload commits run one budget check. The ceiling test seeds `receipt_totals` only. CI is green on this SHA. Codec bounds, signature and window checks, and receipt rollback still look right.

**Blocking:** a second `UploadSession` on the same lease can commit `protocol_done` for receipts it never registered. `record_protocol_done` should require the message sequence to be `received + 1`. The reconnect test should cover `ProtocolDone` as well as a replayed event frame. Details are inline.

**Hold remains.** Keep this unmerged until the archive consumer that lists, satisfies, and compacts `receipts` is reviewed with it. `attempt_progress` still returns only the attempt summary, `MAX_RECEIPTS` still stops admission after two full attempts, and per-event FULL commits are still unbenchmarked. The sequence fix leaves that gate in place.

GitHub reports this PR `CONFLICTING` with master. #45 has merged and creates `user_version` 1; this head only opens version 2 and rejects 1. Rebase onto master before any merge. Rejecting those v1 files matches the undeployed-ledger decision.

PR #48, opened 2026-09-23
Connect isolated reconciliation receipts to durable archive completion

erskingardner:
## Stack and scope
Stacked on #46 (6064c03), which is stacked on #45. Neither prerequisite is merged. This is increment 2a: the actual archive receipt consumer requested by both reviewers, not an enabled worker or production deployment.

## Changes
- Admit registered events through the existing shared dedupe reservation, notepack encoder, and segment writer before returning upload credit; no public ACK-without-admission API.
- ProtocolDone enters awaiting_durability and invalidates the worker capability. Two waiting jobs stop new leases.
- Reconcile at most 256 receipt IDs per call against durable archived markers. Pending and missing IDs never imply completion; a persisted cursor prevents early missing IDs from starving later receipts.
- Atomically compact archive-confirmed receipt detail into attempt summaries, retaining unresolved IDs and all job lineage. Complete only after the current protocol summary and all historical receipt obligations are durable; split parents also require both children.
- Add fenced parent durability retry and adjustable retained receipt capacity. Recovery writes bypass admission ceilings, not actual IO failures.
- Fail closed on incompatible undeployed schema versions rather than silently migrating prototype state.

## Verification
Full just precommit passed immediately before commit: formatting, strict workspace clippy, tests and doctests. Ten new integration tests use real temporary SQLite, RocksDB and notepack files, including an abrupt subprocess exit after sealing, rollback, missing prior-attempt receipts, pending duplicates, reuse beyond the receipt cap, archive failure, deadline expiry, and recovery above the soft admission ceiling. Existing 35 sync tests pass.

## Explicit remaining gates
No runtime scheduler, authenticated socket, worker, SDK fork, service changes, or deployment. Independent max-age sealing and bounded sealed inventory replay remain the next integration slice. Callers must use the same writer/index pair after startup archive recovery and a bounded blocking executor. Per-event FULL-commit throughput remains unbenchmarked on production storage. Job/attempt summaries remain retained under finite admission ceilings. This is not a claim of continuously operable negentropy yet.

See docs/negentropy_archive_receipts.md for contracts and limitations.


crates/pensieve-ingest/src/sync/ipc.rs
```diff
@@ -368,15 +372,74 @@ impl UploadSession {
...
+                let packed = crate::pack_nostr_event(&event.event)?;
+                writer.write_reserved(packed, claim)?;
+            }
+            None => {
+                // Legacy persisted Pending is neither a current writer claim nor
+                // archive proof. Do not acknowledge an owner that does not exist.
+                if dedupe.get_status(event.event.id.as_bytes())?
+                    == Some(crate::EventStatus::Pending)
+                {
+                    return Err(ProtocolError::State);
```
erskingardner: **Serious: this permanently wedges the window.** Builds before #22 wrote `Pending` at claim time and replaced it with `Archived` at seal. A `Pending` still on disk therefore means that build crashed before sealing, so the event is plausibly *missing* from the archive. Negentropy exists to repair exactly this.

With this branch:
- `reserve` returns `None` (because `get_status` returns `Some`), we return `State`, and the session is poisoned.
- `reconcile_archived` only accepts `Archived`, so this receipt is never satisfied, and the job and every ancestor stay incomplete.
- Each retry re-fetches the same ID, fails at the same point, and leaves one more retained receipt that is never compacted. The only exit is to split repeatedly down to `blocked`.

Failing closed is defensible, but the outcome should be deliberate and tested (this branch has no test). Options: admit legacy-`Pending` IDs through a reserve variant that treats them as absent, so they are written and marked `Archived` at seal (the cost is a possible duplicate in the archive); or first run a one-off pass that resolves legacy markers. How many `Pending` values does the production index hold?

crates/pensieve-ingest/src/sync/jobs.rs
```diff
@@ -488,11 +512,157 @@ impl JobLedger {
...
+                    params![job, attempt, sequence],
+                )?;
+                tx.execute(
+                    "UPDATE attempts SET archived=archived+1 WHERE job=?1 AND attempt=?2",
+                    params![job, attempt],
+                )?;
+                satisfied += 1;
+            }
+        }
+        let (next_attempt, next_sequence) = rows.last().map_or((0, 0), |r| (r.0, r.1));
```
erskingardner: Only an *empty* batch wraps the cursor, so every full pass costs an extra no-op transaction. The tests show this (`// wrap the persisted scan cursor`, and `reconcile` is called twice before `.complete`). A natural caller loop like `while r.satisfied > 0` would stop before completion. Wrap as soon as the batch is short:

```rust
let (next_attempt, next_sequence) = match rows.last() {
    Some(r) if rows.len() == limit as usize => (r.0, r.1),
    _ => (0, 0),
};
```

That also removes the extra calls from the tests.

crates/pensieve-ingest/src/sync/jobs.rs
```diff
@@ -488,11 +512,157 @@ impl JobLedger {
...
+fn complete_if_durable(tx: &Transaction<'_>, mut job: i64) -> Result<(), LedgerError> {
+    // An i64-second inclusive interval can split at most 63 times. Bound parent
+    // propagation even in the presence of an invalid externally edited database.
+    for _ in 0..=63 {
+        let changed = tx.execute(
+            "UPDATE jobs SET state='complete',reason=NULL WHERE id=?1
+             AND NOT EXISTS(SELECT 1 FROM receipts WHERE job=?1)
+             AND ((state='awaiting_durability' AND EXISTS(
+                 SELECT 1 FROM attempts WHERE job=?1 AND attempt=jobs.attempt
+                 AND protocol_done=1 AND archived=received))
```
erskingardner: `archived=received` is implied by the `NOT EXISTS(SELECT 1 FROM receipts WHERE job=?1)` above it. Each `received` increment inserts exactly one receipt in the same transaction, and each `archived` increment deletes exactly one. So the counter carries no information that the receipts table does not. Consider deleting the `archived` column and its CHECK, the `UPDATE attempts SET archived=...` at L577, and `AttemptProgress::archived`. If per-attempt reporting is wanted, derive it as `received - count(receipts)`. That is a net deletion and removes one more invariant to keep in sync.

crates/pensieve-ingest/src/sync/ipc.rs
```diff
@@ -368,15 +372,74 @@ impl UploadSession {
...
+        ledger: &mut JobLedger,
+        event: RegisteredEvent,
+        dedupe: &crate::DedupeIndex,
+        writer: &crate::SegmentWriter,
+        mut clock: C,
+    ) -> Result<Accepted, ProtocolError>
+    where
+        C: FnMut() -> i64,
+    {
+        self.check_ack(ledger, &event, clock())?;
```
erskingardner: Nit: each admitted event now runs `check_ack`, including `verify_active` (a SQLite read), twice: once here and again in `accepted`. The safety property is "no ACK after the lease deadline", and the post-admission check already enforces it. Writing a durably registered candidate into the archive after expiry is harmless under archive-first. Similarly, the `recovery_required()` check on L396 duplicates `write_reserved`'s own `archive_operation` gate, which rejects the write and drops the claim unretained. The only extra thing it guards is ACKing a duplicate. Both could go, leaving a single `check_ack` in `accepted`.

crates/pensieve-ingest/src/sync/jobs.rs
```diff
@@ -290,6 +306,14 @@ impl JobLedger {
         if active {
             return Ok(None);
         }
+        let awaiting: u32 = tx.query_row(
+            "SELECT count(*) FROM jobs WHERE state='awaiting_durability'",
+            [],
+            |r| r.get(0),
+        )?;
+        if awaiting >= 2 {
```
erskingardner: Nit: `2` is a policy knob, like `MAX_RECEIPT_BATCH`. Name it (for example `const MAX_AWAITING_DURABILITY: u32 = 2;`) with a one-line rationale. `SELECT count(*) >= ?` or `EXISTS ... LIMIT 1 OFFSET 1` would also avoid counting the whole set, though that cost is negligible here.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-23T20:07:48Z` |
| Commit reviewed | `3b8a2481ded9308e42de3777fb0b96699fecaf76` |
| Reasoning level | `reasoning_effort=15` |

## Summary

This is the consumer that the #46 reviews asked for. `receipts` now has a reader, archive-confirmed detail is compacted, and completion depends on durable `Archived` markers rather than protocol success. That resolves the main premise objection on #46, and the two PRs should be evaluated and merged together. The design is sound: pending and missing IDs never count as proof, obligations from older attempts survive retries, split parents need both children plus their own receipts, and rollback is atomic. The real-storage integration tests (including subprocess exit after seal) are good. CI is green on this SHA.

**Serious concern (inline):** a legacy on-disk `Pending` marker permanently wedges its window. Builds before #22 wrote `Pending` at claim time and overwrote it at seal, so any marker still on disk means that build crashed before sealing. Those events are the ones most likely to be missing from the archive. `admit_and_accept` rejects them and poisons the session, and `reconcile_archived` can never satisfy them. Every retry fails at the same event and leaves one more receipt that is never released. The only way out is to split repeatedly down to `blocked`. This path has no test. It needs an explicit decision (admit them, or migrate or clear the legacy markers) before runtime use.

**Smaller issues (inline):**
1. A short batch does not reset the cursor, so every pass costs an extra no-op call. This is why the tests need `// wrap` calls, and a caller that stops at `satisfied == 0` will never complete the job.
2. The `attempts.archived` counter is redundant. "No receipts remain" already implies `archived = received`, so the column, its CHECK, the per-row UPDATE and the extra predicate can all be deleted.
3. Each admitted event runs `check_ack` (a SQLite read) twice, and the `recovery_required()` pre-check duplicates the writer's own admission gate.
4. `awaiting >= 2` is a magic number; give it a named const next to `MAX_RECEIPT_BATCH`.

The remaining gates the PR states (max-age sealing, scheduler, transport, throughput) are the right ones. Without max-age sealing, two awaiting jobs can stall leasing until 256 MB of live traffic arrives. That is already in the doc and does not block this library increment.

crates/pensieve-ingest/src/sync/ipc.rs
```diff
@@ -0,0 +1,877 @@
...
+        self.closed = true;
+        if writer.recovery_required() {
+            return Err(ProtocolError::State);
+        }
+        match dedupe.reserve(event.event.id.as_bytes())? {
+            Some(claim) => {
+                let packed = crate::pack_nostr_event(&event.event)?;
+                writer.write_reserved(packed, claim)?;
+            }
+            None => {
```
erskingardner: `reserve` returns `None` for both a durable row and an in-memory claim. Disk `Pending` correctly fails closed here. In-memory pending does not: `get_status` only reads RocksDB, so it is `None`, and this arm falls through to `accepted`.

That is the right outcome for a second frame of an ID `write_reserved` already retained in this writer. Seal will mark it `Archived`. It is the wrong outcome when live ingest has only reserved the ID. The negentropy handler in `main.rs` drops that `PendingAdmission` without `retain()` if packing fails, and a crash before `write_reserved` drops the in-memory set with no segment bytes. This ACK still lets `ProtocolDone` move the job to `awaiting_durability`, while `reconcile_archived` will never observe `Archived`. `expire` does not apply to that state, so the job holds one of the two lease slots until something calls `retry_durability`.

ACK this arm only when status is `Archived`, or when this writer already has the ID in the open segment. A foreign claim that has not retained bytes should stay `State`, with the receipt preserved.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-23T20:14:29Z` |
| Commit reviewed | `3b8a2481ded9308e42de3777fb0b96699fecaf76` |
| Model | `Grok 4.7` |

## Summary

The consumer matches the contract from #46: credit is released only through `admit_and_accept`, `ProtocolDone` is not completion, reconciliation trusts `Archived` markers only, and a missing early ID cannot pin the scan cursor. Compaction, split-parent completion, and the crash-after-seal test are atomic or a real reopen. CI is green on this SHA.

The shared-index ACK still treats every in-memory pending ID as safe to acknowledge. That holds for a duplicate frame this writer already stored, and not for a live-ingest claim that is later dropped. Details are on `admit_and_accept`.

This PR merges cleanly into `codex/negentropy-ipc-receipts`. That branch (#46) is currently conflicting with master, and it introduces `user_version` 2 only for this PR to reject it. Rebase #46 first. If the two increments land as one schema, keep a single `user_version` instead of shipping a version that never exists outside the stack.

erskingardner:
Addressed both current-round serious findings in 2933b19 (rebased on corrected #46). Legacy on-disk Pending is re-admitted without deleting the marker, under the shared admission mutex; only a new durable seal establishes Archived. Revocable live-source reservations are distinguished from retained writer obligations and cannot ACK. New real-storage regressions cover dropped claims, retries preserving old receipts, legacy recovery, duplicate uploads and seal proof. The short-batch cursor now wraps immediately and the awaiting limit is named. Full just precommit passed.

Retained the archived-frame summary because per-attempt accounting is part of the approved product contract. Retained pre/post lease checks intentionally: reject invalid handoffs before work, and reject ACK after slow admission crosses expiry. Retained the recovery precheck because duplicate admission otherwise bypasses the writer. Prototype schema identifiers remain distinct: previously created v2 files must fail closed, not be silently reinterpreted as the different v3 layout. Neither schema has a production runtime caller; no deployment is authorized by merging these library increments.

crates/pensieve-ingest/src/sync/jobs.rs
```diff
@@ -492,11 +518,162 @@ impl JobLedger {
...
+            checked: rows.len() as u32,
+            satisfied,
+            complete,
+        })
+    }
+
+    /// Explicit parent recovery when a finished upload cannot reach durability.
+    /// Preserves every receipt and retries the same interval with a new capability.
+    /// The attempt number fences stale recovery decisions. Not a worker message.
+    pub fn retry_durability(
```
erskingardner: **Non-blocking: an unsatisfiable receipt has no terminal outcome.** Suppose attempt N registers event X and the session then dies before admission (crash, a lost foreign-claim race, or the deadline passing during a slow admission). Later the relay drops X (NIP-09 deletion, NIP-40 expiry, or a retention policy). Every later attempt finishes `ProtocolDone` without X and enters `awaiting_durability`. `reconcile_archived` can never satisfy X's receipt, so the only exit is this function, which retries the same window with the same result. The job cycles forever, and each cycle holds one of the `MAX_AWAITING_DURABILITY` slots. For a split job, one old receipt keeps the parent and every ancestor in `split` even after both children complete. Splitting cannot reach `blocked` for this case either.

Failing closed on coverage is correct. What is missing is a named terminal state, for example `blocked` with reason `unrecoverable_receipt` after N durability retries, or an explicit operator resolution API. Without one, the future scheduler churns instead of surfacing the job. This does not need to be fixed in this increment. Please add it to the "Still required" list in `docs/negentropy_archive_receipts.md`.

crates/pensieve-ingest/tests/sync_archive_receipts.rs
```diff
@@ -0,0 +1,647 @@
...
+    let mut session = UploadSession::new(lease.clone());
+    for seq in 1..=2 {
+        let received = h.receive(&mut session, &lease, seq, &candidate, 11);
+        h.admit(&mut session, received, 12);
+    }
+    assert!(h.dedupe.has_archive_owner(candidate.id.as_bytes()).unwrap());
+    h.done(&mut session, &lease, 13);
+    assert!(!h.reconcile(lease.job().id, 4).complete);
+    assert_eq!(h.writer.seal().unwrap().unwrap().event_count, 1);
+    h.reconcile(lease.job().id, 4);
```
erskingardner: Nit: this call is no longer needed. Two receipts with `limit = 4` form a short batch, so the call on L182 already wrapped the cursor, and the next call satisfies both rows and completes the job. The extra call suggests that callers must always reconcile twice. The same applies to L231 in `foreign_revocable_claim_cannot_ack_and_dropped_claim_is_recoverable`. (L309 is different: there the batch was exactly full, so the extra call is legitimately needed.)

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-23T20:25:45Z` |
| Commit reviewed | `2933b19509efa23216e3b02ff7a3b0a5383fadac` |
| Reasoning level | `reasoning_effort=15` |

## Summary

Re-review at 2933b19. Both serious findings from the previous round are fixed and tested against real storage:

- **Legacy on-disk `Pending`:** `reserve_unarchived` re-admits the ID under the shared admission mutex and leaves the marker in place. Only a new durable seal writes `Archived`. `legacy_pending_is_readmitted_without_inventing_archive_proof` covers a dropped claim, a duplicate frame, and the seal proof.
- **Foreign revocable claim:** `pending` now distinguishes a revocable reservation (`false`) from a retained writer obligation (`true`). `has_archive_owner` ACKs only a retained claim or an `Archived` marker, so a live-source reservation that has not written bytes cannot ACK. `foreign_revocable_claim_cannot_ack_and_dropped_claim_is_recoverable` covers this, including recovery on the next attempt.
- A short batch now wraps the cursor, and the awaiting limit is named `MAX_AWAITING_DURABILITY`.

The author gave reasons for keeping the `attempts.archived` counter (per-attempt accounting is part of the product contract) and the pre- and post-admission lease checks. I accept both. #46 is now rebased and mergeable. CI is green on this SHA (clippy, doctests, rustfmt, all three test shards). These two PRs should land together.

**Non-blocking (inline):**
1. A receipt whose event disappears from the relay has no terminal outcome. Its job loops between `awaiting_durability` and `retry_durability` forever, and a split parent holding such a receipt can never complete. That is acceptable for this library increment, but add it to the documented remaining gates before a scheduler drives these APIs.
2. Two tests keep a `reconcile` call that the short-batch wrap made unnecessary.

Posted as a comment rather than an approval because GitHub does not let the PR author's account approve its own PR. The recommendation is **Merge**.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-23T20:27:01Z` |
| Commit reviewed | `2933b19509efa23216e3b02ff7a3b0a5383fadac` |
| Model | `Grok 4.7` |

## Summary

The previous serious findings are fixed on this head, and CI is green (fmt, clippy, doctests, and all three test shards).

Legacy on-disk `Pending` is re-admitted through `reserve_unarchived` without deleting the marker, and only a later seal writes `Archived`. A revocable live claim cannot ACK; credit is released only for a retained writer claim or an `Archived` marker. A short or empty reconciliation batch resets the scan cursor. `MAX_AWAITING_DURABILITY` names the lease pause. The new real-storage tests cover dropped claims, legacy recovery, duplicate uploads, and seal proof.

Completion still requires a checked protocol summary and no remaining receipt rows. Pending and missing IDs never count. The retained `archived` counter cannot by itself complete a job while any receipt row remains, and the only delete path increments it in the same transaction. Keeping the pre- and post-admission lease checks matches the slow-admission case the tests lock in.

The stated gates still belong to the next slice: max-age sealing, a scheduler that keeps reconciliation running while leases are paused, transport, and throughput. Two unsealed `awaiting_durability` jobs will stop new leases until something seals or calls `retry_durability`. That is already the contract in `docs/negentropy_archive_receipts.md`, and this increment does not enable a runtime caller. Land it with #46.

GitHub rejected `--approve` because this account opened the pull request. The recommendation is still to merge.

PR #49, opened 2026-09-23
Seal the archive independently of optional Parquet publication

erskingardner:
## Scope
Next isolated-negentropy slice, stacked on #48. Periodic archive sealing no longer depends on successful Parquet startup. Adds a configurable 300-second archive cadence; active Parquet may request a shorter interval. Both zero values disable periodic sealing explicitly.

One owned background thread performs synchronous seals with no work queue or catch-up burst. Shutdown wakes and joins it before the final seal; archive failure stops intake and preserves the existing recovery latch. No worker, IPC listener, scheduler, deployment or production restart.

## Verification
Full just precommit passed, including after rebasing onto the reviewed receipt corrections. Tests cover Parquet-disabled durable sealing, archive failure and preserved evidence, idle shutdown, joining in-flight work, and configuration precedence.

Includes docs/negentropy_execution_plan.md as the tracked approved plan and remaining gates. Bounded sealed inventory replay is next; Linux canary and actual rollout remain separate gates.


crates/pensieve-ingest/src/main.rs
```diff
@@ -229,6 +230,15 @@ struct Args {
     )]
     parquet_shadow_max_batch_age_secs: u64,
 
+    /// Periodically seal the authoritative archive, independently of Parquet.
+    /// Zero disables this timer; isolated reconciliation requires a nonzero cadence.
```
erskingardner: `--help` still describes a single timer with a single off switch.

This line says zero disables this timer. A few lines above, `--parquet-shadow-max-batch-age-secs` still says "Zero disables the timer." `archive_seal_interval` starts a thread whenever either positive value applies, and an active shadow keeps the parquet age even when the archive interval is zero (`archive_cadence_is_independent_of_optional_shadow`).

Please match both doc comments to the README: each zero removes only that input, the period is the minimum of the remaining positive values, and periodic sealing stops only when both are zero.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-23T20:27:31Z` |
| Commit reviewed | `46fe8463673a7d673ea4874a0b63df417aa1f748` |
| Model | `Grok 4.7` |

## Summary

This is the max-age seal #48 needs so receipt durability does not wait for a 256MB segment. One owned thread, no tick queue, and shutdown joins that thread before the final seal. `seal()` stays on the existing admission lock, so a timer seal and a size-threshold seal cannot publish the same segment number. An I/O error latches recovery, stops intake, and leaves the resident recovery wait in place.

The cadence rule (minimum of the positive intervals; either zero drops only that input) matches the new test. Idle ticks seal nothing when no segment is open.

rustfmt, clippy, doctests, and test shard 3/3 were green on this SHA. Test shards 1 and 2 were still running; I did not re-run the suite locally.

GitHub rejected an APPROVE event because this review is from the PR author. The recommendation is still to merge.

crates/pensieve-ingest/src/main.rs
erskingardner: Related stale doc a few lines up (line 225, outside the diff): the `parquet_shadow_max_batch_age_secs` comment still says "Zero disables the timer." That is no longer true. Zero only removes the Parquet request, and this flag (default 300) keeps sealing. Since `archive_seal_interval` takes the `min`, any existing deployment that set the Parquet age above 300 to get larger objects will now get 300s batches with no warning. Suggest rewording that comment to: "Requests a seal cadence no longer than this while the shadow is active; the effective interval is the minimum of this and `--archive-seal-interval-secs`. Zero withdraws the request."

README.md
```diff
@@ -77,6 +77,15 @@ Connect to Nostr relays via WebSocket and stream events in real-time. Includes a
 - Disconnects from paid/whitelist relays that reject auth
 - **Relay quality tracking** (SQLite) with scoring and optimization
 - Optional Parquet shadow publication from durably sealed notepack work units
+
+Live archive segments are periodically sealed independently of Parquet. Configure
```
erskingardner: This paragraph sits between two bullets under **Features:**. That ends the list, so `- Graceful shutdown on Ctrl+C` renders as a separate one-item list below it. Move the paragraph after the list, or shorten it to a sub-bullet. The flag should probably also get a row in the Options table below, next to the other ingest flags.

crates/pensieve-ingest/src/pipeline/seal_timer.rs
```diff
@@ -0,0 +1,222 @@
...
+            true
+        })
+        .unwrap();
+        ready.recv_timeout(Duration::from_secs(5)).unwrap();
+        let (finished, done) = mpsc::sync_channel(1);
+        let joiner = thread::spawn(move || {
+            timer.shutdown().unwrap();
+            finished.send(()).unwrap();
+        });
+        assert!(done.recv_timeout(Duration::from_millis(25)).is_err());
```
erskingardner: This could flake on a busy CI runner. The test assumes the `joiner` thread has run `try_send(())` within 25ms. If it hasn't, `release` lets the first seal return, `recv_timeout(1ms)` times out, and a second seal starts and blocks forever on `wait.recv()`, because `release` stays alive and never sends again. The join then hangs and the 5s `done.recv_timeout` fails. A fix that doesn't depend on timing: send the stop signal before releasing, e.g. `drop(release)` after `shutdown` has certainly started, or have the closure return `false` after the first call. Either way the test still checks that shutdown waited for the in-flight seal.

docs/negentropy_execution_plan.md
```diff
@@ -0,0 +1,93 @@
...
+- Uptime Kuma is the chosen alert destination. Health stays failed until the actual
+  recovery condition resolves. Verify delivery during rollout; an elaborate universal
+  recovery procedure is not required.
+
+## Reviewable slices and current status
+
+| Slice | Scope | Gate / status |
+| --- | --- | --- |
+| 1a, #45 | Durable jobs, leases, retry and split | Merged; CI and review passed |
+| 1b, #46 | Bounded framed uploads, credits, received receipts | Review/CI gate; library only |
```
erskingardner: Scope nit. This 93-line plan isn't needed for the seal-cadence change. The status column ("Review/CI gate", "Current implementation", "Next") will be wrong as soon as #46, #48, or this PR merges, and then every later slice has to fix it. Consider landing the plan on its own, or dropping the status column and linking the PRs/issue instead so the doc only states the stable contracts.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-23T20:31:27Z` |
| Commit reviewed | `46fe8463673a7d673ea4874a0b63df417aa1f748` |

(Posted as a comment because GitHub does not let the author approve their own PR. The recommendation is to merge after #48.)

**Premise:** Sound. The periodic seal used to be tied to `parquet_shadow_handle.is_some()`, so a shadow that failed startup, which the RUNBOOK treats as a supported degraded mode, also turned off time-based archive sealing. Decoupling the two is the right fix at the root, and receipt reconciliation needs it. The dedicated thread also takes the synchronous `seal()` (fsync, rename, marker writes) off the tokio reactor, which the old task blocked. One note on the stated motivation: `abort()` on the old task could not interrupt an in-flight synchronous `seal()` either, since the task had no await point inside it. So the "join, don't abandon" rule was already met by accident. The real gains are reactor isolation and making that rule explicit.

**Correctness:** I found no blocking issues.
- Every `seal()` error goes through `archive_operation`, which latches `recovery_required` before the timer's `running.store(false)`. The fault watch and the final-seal guard therefore both see the latch, and shutdown lands in `wait_for_archive_recovery` as intended.
- Timer ticks while idle are cheap, because `seal_checkpointed` returns early when no segment is open.
- On early `?` returns in `main`, `Drop` still wakes and joins the thread.
- Production already runs the shadow at 300s, so the new default should not change segment cadence there.

**Non-blocking (see inline comments):**
- The Parquet max-age flag's doc comment is now wrong. It also doesn't mention that the archive default now caps any Parquet age above 300s.
- The README paragraph splits the Features bullet list.
- One timer test depends on a 25ms scheduling window and could flake.
- The execution-plan doc's status table will be out of date as soon as this stack merges.

CI is green at this head: clippy, rustfmt, doctests, and all three test shards.

erskingardner:
Both reviews recommend merge. Addressed the actionable non-blocking feedback: CLI help now states minimum-positive cadence and independent zero semantics; README list and options row corrected; the shutdown test enqueues stop before releasing the blocked callback so scheduling cannot start a second seal. Full just precommit passed immediately before this fix commit. The tracked execution plan is intentionally included because the user requested a handy plan; it will remain a maintained checklist. No production deployment.

PR #50, opened 2026-09-23
Build bounded archive-confirmed reconciliation inventory

erskingardner:
## Scope
Increment 2c, stacked on #49. Library only: no runtime replay, production scan, worker, scheduler or deployment.

Export fixed inclusive windows with at most 250,000 examined IDs, verifying durable archive markers and rejecting overflow without a truncated result. Stream exactly the next sealed plain/gzip segment in steps of at most 256 frames with bounded payload decoding. Persist explicit archive/prefix/floor identity and advance the cursor only after inventory WAL synchronization. Missing segments, malformed input and unarchived IDs fail closed; unfinished segments replay idempotently after crashes.

## Verification
Full just precommit passed immediately before commit. Tests cover range/overflow, partial replay/reopen, abrupt process exit before and after completion, gzip, missing segments, configuration drift, concurrent reader exclusion, malformed frames and absent durable markers.

See docs/negentropy_archived_inventory.md for bounds and integration limitations. Runtime executor ownership, retained-history protection, scheduler, peer-authenticated IPC, worker isolation and canary remain separate gates.


crates/pensieve-ingest/src/sync/inventory.rs
```diff
@@ -0,0 +1,484 @@
...
+                    "replay frame exceeds bounded decoder".to_owned(),
+                ));
+            }
+            let mut payload = vec![0; length];
+            self.reader.read_exact(&mut payload)?;
+            let event = pensieve_parquet::CanonicalEvent::from_notepack(&payload)
+                .map_err(|e| Error::Validation(e.to_string()))?;
+            if self.dedupe.get_status(event.id())? != Some(EventStatus::Archived) {
+                return Err(Error::Validation(
+                    "sealed replay ID lacks durable archive marker".to_owned(),
```
erskingardner: Seal race: `SegmentWriter::seal_inner` renames `.open` to `.notepack` and fsyncs the directory **before** it calls `dedupe.mark_archived`. The window is therefore the directory fsync plus a RocksDB batch, and on the HDD archive that can be tens of ms. A `begin` that runs during it opens the new segment, and the first `step` fails with this `Validation` error. The session is poisoned, and the error looks exactly like a corrupt or unarchived segment.

This fails closed, so no data is lost. But a polling scheduler will hit it now and then in normal operation, and it can't separate this case from real corruption. Two options:
- have `begin` return `Ok(None)` until the writer has finished marking the segment, e.g. using the writer's sealed watermark/segment number, or
- return a distinct retryable error for this case.

crates/pensieve-ingest/src/sync/inventory.rs
```diff
@@ -0,0 +1,484 @@
...
+    ///
+    /// None means the next segment has not sealed yet. A higher sealed segment
+    /// with a missing predecessor is an error, never an instruction to skip it.
+    /// Run after archive startup recovery, on a bounded blocking executor. Do not
+    /// run the legacy inventory pruner while replay or retained jobs need history.
+    pub fn begin(
+        state: &'a SyncStateDb,
+        dedupe: &'a DedupeIndex,
+        writer: &'a SegmentWriter,
+        archive: &Path,
```
erskingardner: `begin` already takes `&SegmentWriter`, but it also takes `archive` and `prefix` as separate arguments. Nothing checks that they match the writer's `output_dir`/`segment_prefix`. A caller can therefore replay one directory while gating on a different archive's `recovery_required()`.

A small `pub(crate)` accessor on `SegmentWriter` for its config would let you take both values from the writer. That drops two parameters and the prefix-validation block (the writer already owns that name). `segment_number` below also duplicates the suffix parsing in `SegmentWriter::find_next_segment_number`, so the two could share one helper.

crates/pensieve-ingest/src/sync/inventory.rs
```diff
@@ -0,0 +1,484 @@
...
+            }
+            batch.push((*event.id(), event.created_at()));
+        }
+        if self.writer.recovery_required() {
+            return Err(Error::Validation("archive recovery required".to_owned()));
+        }
+        self.state
+            .record_batch(batch.iter().map(|(id, timestamp)| (id, *timestamp)))?;
+        if eof {
+            // Refuse concurrent/stale readers rather than moving a cursor backward.
```
erskingardner: I think this re-read and compare can't fail. `replay_lock` is held for the whole lifetime of `InventoryReplay`, and RocksDB's `LOCK` file stops a second process or handle from opening the same sync DB. So nothing else can write the cursor between `begin` and here. Suggest deleting these ~10 lines. The same applies to the `cursor.next < floor` check in `begin`: this code only ever writes `next >= floor`.

crates/pensieve-ingest/src/sync/state.rs
```diff
@@ -42,9 +56,81 @@ use std::path::Path;
...
+        Ok(cursor)
+    }
+
+    pub(super) fn save_replay_cursor(&self, bytes: &[u8]) -> Result<()> {
+        if bytes.len() > 8192 {
+            return Err(crate::Error::Config("oversized replay cursor".to_owned()));
+        }
+        // All earlier inventory puts reach stable WAL before its completion cursor.
+        // A crash in between repeats valid puts, never skips unflushed inventory.
+        self.db.flush_wal(true)?;
```
erskingardner: `flush_wal(true)` only matters when `manual_wal_flush` is enabled, and it isn't set here. Without it, `put_opt` with `set_sync(true)` fsyncs the WAL, and that covers every earlier write already appended to it (older writes flushed to SSTs are already durable). The synced put alone gives the ordering this comment describes.

The 8192-byte caps in `replay_cursor`/`save_replay_cursor` are also speculative, since this module is the only thing that produces the value. Dropping both would make these helpers two lines each.

crates/pensieve-ingest/src/sync/state.rs
```diff
@@ -42,9 +56,81 @@ use std::path::Path;
...
+            return Err(crate::Error::Config(
+                "invalid inventory window bounds".to_owned(),
+            ));
+        }
+        let start = Self::make_key(since, &[0; 32]);
+        let end = Self::make_key(until, &[u8::MAX; 32]);
+        let mut items = Vec::new();
+        for (examined, item) in self
+            .db
+            .iterator(IteratorMode::From(&start, rocksdb::Direction::Forward))
```
erskingardner: This DB is opened with a fixed 8-byte prefix extractor, and this scan crosses prefixes whenever `since < until`. With default `ReadOptions` (`total_order_seek = false`), RocksDB only guarantees correct results past the seek prefix when there are no prefix bloom filters or hash memtables. That holds today, but only implicitly.

Because `Complete` claims to be the full set, suggest using `iterator_opt` with `ReadOptions::set_total_order_seek(true)`, so a future tuning change can't silently drop IDs from a window. `get_items_since` has the same latent assumption.

docs/negentropy_archived_inventory.md
```diff
@@ -0,0 +1,46 @@
...
+`SyncStateDb::archived_window` exports an inclusive fixed interval in timestamp/ID
+order. It checks durable Archived markers for legacy entries; Pending and missing
+IDs are omitted, allowing redundant downloads rather than suppressing recovery.
+Both examined and returned rows are capped at 250,000. TooDense returns **no usable
+partial set**; the scheduler must split the interval or block/alert at one second.
+This set describes known archive inventory, not complete relay history.
+
+`InventoryReplay::begin` binds a canonical archive directory, prefix and explicit
+rollout floor to a versioned cursor in the existing sync RocksDB. There is no
+default history floor. Later configuration changes fail closed. Only one replay
```
erskingardner: Refusing a changed configuration is the right default. But the cursor records the canonical absolute path and the floor, and there is no reset API. So a mistyped floor on first `begin`, or a remount of the archive at a new path, blocks replay permanently. The only way out today is deleting the whole sync DB, which also throws away inventory. It would help to document the intended operator recovery here, or add a narrow explicit reset, before the scheduler gate.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-23T20:39:28Z` |
| Commit reviewed | `1e39dfd37f86d5c220b2d5d844afa75967d0bd8d` |

This is a sound 2c slice. Replay only moves forward on a durable EOF, a partial segment is re-read after a restart, the reserved cursor key sorts after the pruner's end key and after every window end, and the TooDense bound is exact at `max_items`. I did not find any path that loses data or skips a segment. The comments below are non-blocking and mostly about the API the future scheduler will use.

- **Seal race:** a segment is visible under its sealed name before `mark_archived`. Replay can open it during that gap and fail as if the segment were corrupt. The scheduler needs a way to tell "not ready yet" apart from real corruption.
- **Configuration duplication:** `begin` takes `archive`/`prefix` separately from the `SegmentWriter` whose recovery latch it checks. Getting them from the writer removes a whole mismatch class and the prefix validation block.
- **Unneeded defensive code:** the stale-cursor re-read, the 8 KiB cursor caps and `flush_wal(true)` guard against things that can't happen here, or that the synced put already covers.
- **Iterator hardening:** set `total_order_seek` on the new range scan, since a prefix extractor is configured.
- **Operator recovery:** a cursor that fails closed has no reset path.

**CI:** `test (shard 2/3)` failed during `git` checkout with a TLS certificate error, before any tests ran. That is an infrastructure failure, not a code failure. Rerun it to get a green `CI success` before merging. The other shards, clippy, rustfmt and doctests pass.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-23T20:42:11Z` |
| Commit reviewed | `1e39dfd37f86d5c220b2d5d844afa75967d0bd8d` |
| Model | `Grok 4.7` |

## Summary

Increment 2c matches its own contract: library-only, fixed windows, and fail-closed replay. I would merge it. GitHub rejected `--approve` because this is your own pull request, so this is posted as a comment.

`archived_window` either returns every in-range key it was willing to examine or `TooDense` with no partial set. Pending and missing dedupe rows are omitted and still count toward that budget. The 41-byte cursor sorts after every 40-byte timestamp/ID key, so `get_items_since` skips it and `prune_before` cannot delete it. That holds through `u64::MAX`, and the new prune test covers the extreme.

Replay opens only the next sealed plain or gzip name. `.open` and `.gz.open` files are invisible, so an unsealed successor is not a gap and a missing predecessor is not skipped. Preferring the plain file matches the compressor: gzip is renamed into place only after it is complete, and the plain file is removed only after that. Cursor advancement happens after `flush_wal(true)` and a sync put, and only when this session's cursor is still current. A crash before that repeats the segment; `record_batch` is an idempotent put.

Two caller contracts are already written down and should stay intact when the scheduler lands. `replay_lock` does not cover `prune_before`, so the legacy pruner can still delete replayed history and the cursor will not bring it back. Non-final steps are visible to `archived_window` before `flush_wal`; only the EOF path syncs the WAL before moving the cursor. `ArchivedWindow::Complete` means the scan finished inside the budget, not that the relay range is covered.

CI on this SHA is not fully green: shard 2 died in `actions/checkout` (`server certificate verification failed`) before tests ran. Clippy, rustfmt, doctests, and shard 1 passed. I did not re-run the suite locally.

erskingardner:
Both independent reviewers recommended Merge on 1e39dfd. Addressed the cross-prefix iterator concern with explicit total_order_seek and documented the remaining runtime activation gates (seal-in-progress distinction, writer-bound source identity, and non-destructive cursor repair). Kept the small durable-cursor validation checks; no production replay is enabled. Full just precommit passed before the fix commit. Rebased onto merged #49; waiting for fresh CI on f21b2a4 before merging.

PR #51, opened 2026-09-24
Add isolated one-job negentropy worker session

erskingardner:
## Scope
Increment 3a of the approved process-isolation plan: an experimental one-job worker executable, bounded parent inventory/ACK protocol, Unix parent-UID verification, backpressured wire capture, full lifecycle/idle/lease deadlines, and explicit diff/fetch/capture/ACK completion checks. No SDK fork or version upgrade.

The worker uses SDK relay transport but owns its download-only NIP-77 loop. It does not use SDK sync_with_items, which can report success after notification loss. Lag/closure fails closed; explicit diff completion and fully fetched 128-ID batches are required. Inventory and missing-ID sets are capped; duplicate inventory IDs at different timestamps are rejected.

## Validation
Ten focused tests pass, including a localhost NIP-77 relay, genuine empty diff, partial fetch despite EOSE, notification lag/closure, hung relay, parent disconnect, stalled inventory, wrong UID, malformed inventory, bounded queue/byte backpressure, and lost/wrong ACK retaining received obligations. Full just precommit passed immediately before each commit. Initial-head CI passed; fresh CI is required on the correction commit.

## Review gate
Grok identified two blockers on the initial head; both are addressed in 78b4110. The corrected completion path still needs independent review. Claude exited before reviewing because CLI OAuth authentication expired. Do not treat this as a passed review or merge yet.

## Boundaries
No production deployment or ingester runtime wiring. Parent scheduler/authenticated listener/inventory activation remain next; Linux cgroup limits, permissions, failure injection and real alert delivery remain mandatory before enabling. Process exit is worker cleanup; asynchronous deadlines alone cannot contain non-yielding code or all SDK decoder OOM. Tracked execution plan and worker contract are updated.


erskingardner:
Full local precommit and the standalone worker --help smoke test passed. Independent review dispatch: Grok 4.7 Fast is running; Claude Opus 5.5 exited 1 before reviewing because its OAuth session expired and could not be refreshed. This is a missing review, not an approval. Waiting for restored CLI authentication and current-head CI before merge.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-24T07:02:26Z` |
| Commit reviewed | `a3239c7e229a512722a69aff3ef76fc744992f2d` |
| Model | `Grok 4.7` |

The session matches the increment 3a boundary: one job, UID check before Hello, credit held until a matching admission ACK, and ProtocolDone only after the upload drains. Exact-head CI is green. It still must not land while an aborted SDK sync can become a zero-event ProtocolDone and a completed window.

`nostr-relay-pool` 0.44.3 `InnerRelay::sync` returns `Ok(())` when its broadcast receiver lags or closes, and `finish` treats the resulting empty `remote` set as a complete diff. A second gap: inventory accepts the same event ID at a later timestamp.

crates/pensieve-ingest/src/sync/worker/capture.rs
```diff
@@ -0,0 +1,307 @@
...
+    }
+
+    pub async fn finish(
+        &self,
+        result: Option<&Reconciliation>,
+    ) -> Result<(u64, [u8; 32]), WorkerError> {
+        let mut state = self.state.lock().await;
+        state.sender.take();
+        let result = result.ok_or(WorkerError::Incomplete)?;
+        if state.failed
```
erskingardner: `finish` treats `Ok(Reconciliation)` as a finished diff. In `nostr-relay-pool` 0.44.3, `InnerRelay::sync` is `while let Ok(notification) = notifications.recv().await`, then falls through to `Ok(())`. `broadcast::RecvError::Lagged` and `Closed` are both `Err`, so that loop stops without `sync_done`.

`notifications` is subscribed before `check_negentropy_support` and is not polled until that check returns. The check uses a second subscriber and waits up to the default 10s. The unread buffer is 2048 (`DEFAULT_NOTIFICATION_CHANNEL_SIZE`). `handle_relay_message` broadcasts every decoded relay frame. Once more than 2048 frames have been sent, the first `recv` on the unread subscriber returns `Lagged`, and `sync_with_items` still returns `Ok` with the default empty sets.

Empty `remote` is a subset of both `received` and `state.ids`, so this `Ok` passes. Frames already accepted by `save_event` stay in the digest. `record_protocol_done` with `archived = received` (including both zero) satisfies `complete_if_durable`, and the window can be marked complete without a finished reconciliation.

Do not send ProtocolDone on this `Ok` until a lagged or closed notification loop is an error. Run that loop in the worker, where that exit is `Incomplete`, or pin an SDK build that returns `Err` unless `sync_done` is set. The localhost EOSE test does not cover this: the support-check subscriber is a different receiver from the one that falls through.

crates/pensieve-ingest/src/sync/worker/transport.rs
```diff
@@ -0,0 +1,241 @@
...
+                    || chunk.is_empty()
+                    || chunk.len() > INVENTORY_CHUNK_ITEMS
+                    || items.len() + chunk.len() > MAX_WINDOW_ITEMS
+                {
+                    return Err(ProtocolError::Limit);
+                }
+                for (id, timestamp) in chunk {
+                    if timestamp < assignment.since
+                        || timestamp > assignment.until
+                        || previous.is_some_and(|p| p >= (timestamp, id))
```
erskingardner: The worker contract says inventory rejects duplicates. `previous >= (timestamp, id)` rejects an exact pair and any out-of-order pair. The same 32-byte ID at a later timestamp is still accepted.

`NegentropyStorageVector::seal` only `dedup`s equal items, so two timestamps for one ID both remain in the fingerprint. Reject a repeated ID even when its timestamp moves forward.

erskingardner:
Addressed both Grok findings in a separate fix commit. The worker no longer calls the SDK convenience sync loop. Pensieve owns a download-only NIP-77 loop over SDK relay transport, with one notification receiver: Lagged/Closed are explicit errors; only engine diff completion plus fully received 128-ID batches constructs download proof. Missing IDs are capped at 50,000. No SDK fork or version upgrade. Inventory now rejects the same ID at different timestamps.

Ten focused tests pass, including notification lag/closure, real empty diff, partial fetch despite EOSE, wrong/lost ACK preserving receipts, and duplicate-ID regression. Final full just precommit passed immediately before the fix commit. Fresh CI is required on the new head.

Merge remains blocked on independent review of this corrected path. Claude did not review because its CLI OAuth session expired; user has been asked to restore authentication. Both review CLI processes have exited and been reaped. No production changes.

crates/pensieve-ingest/src/sync/worker.rs
```diff
@@ -0,0 +1,166 @@
...
+
+async fn attempt(
+    mut stream: UnixStream,
+    assignment: Assignment,
+    items: Vec<(EventId, Timestamp)>,
+    idle_time: Duration,
+) -> Result<(), WorkerError> {
+    let (capture, mut events) = Capture::new(&assignment);
+    let capture = Arc::new(capture);
+    let client = Client::builder().database(capture.clone()).build();
```
erskingardner: **Serious: some advertised IDs can never be captured, so the window fails on every attempt.**

In `nostr-relay-pool` 0.44.3, `handle_event_msg` drops some events before it calls `save_event` or sends any notification:

- more than 2,000 tags (10,000 for kind 3), from the default `RelayEventLimits`
- `event.is_expired()`, which is unconditional for NIP-40

After that, `Capture` rejects any event whose frame is over 1 MiB, while the SDK accepts messages up to 5 MB.

Negentropy doesn't know about any of this, so these IDs still show up in `remote`. The batch EOSE check then fails. The failure repeats on every attempt, and the parent splits the window until it reaches `blocked`.

The tag limit is the worst case. `source/relay.rs` uses the same default limits, so live ingest never archives events with more than 2,000 tags. The parent inventory will never contain them, so every relay window that has one fails.

Suggestions:

- Build this relay with `RelayLimits` that match archive admission. `RelayLimits::disable()`, or at least no tag cap, would work, since `validate_archive_event` only checks id and signature.
- Decide how to handle IDs that can't be recovered (expired, or over the frame limit). Return a distinct error so the parent doesn't spend retries and splits on a failure that will repeat forever.
- Add a fixture test with a 2,001-tag event.

crates/pensieve-ingest/src/sync/worker/reconcile.rs
```diff
@@ -0,0 +1,193 @@
...
+    loop {
+        match notification(&mut notifications).await? {
+            RelayNotification::Message {
+                message:
+                    RelayMessage::NegMsg {
+                        subscription_id,
+                        message,
+                    },
+            } if subscription_id.as_ref() == &id => {
+                if message.len() > NEG_FRAME * 2 {
```
erskingardner: **The NEG-MSG size limit is below what real relays send.**

`NEG_FRAME` limits the frames *we* send. NIP-77 doesn't limit the responder by our frame size. strfry uses `Negentropy ne(storage, 500'000)` (`RelayNegentropy.cpp:99`) and khatru uses `negentropy.New(vec, 1024*1024)`. As the range count grows each round, their replies grow up to those limits. That happens once roughly 1,900 IDs are missing in one round (60 KB / 32 B).

This check turns those diffs into `Incomplete`, and they are exactly the high-gap windows this worker is for. Splitting eventually works around it, but dense seconds end up `blocked`.

The SDK's 5 MB `max_size` message limit already bounds allocation before we get here. I'd drop this check, or raise it to that limit.

crates/pensieve-ingest/src/sync/worker/reconcile.rs
```diff
@@ -0,0 +1,193 @@
...
+
+use super::super::ipc::MAX_EVENTS;
+use super::{Assignment, WorkerError};
+
+const NEG_FRAME: usize = 60_000;
+const FETCH_BATCH: usize = 128;
+
+/// Constructed only after a complete diff and every requested fetch reaches EOSE.
+#[cfg_attr(test, derive(Default))]
+pub(super) struct DownloadProof {
```
erskingardner: Nit/simplification: `received` is redundant. `download` returns `Ok` only after every batch passes `batch.iter().all(|id| received.contains(id))`, and every ID in `remote` is in some batch. So `remote ⊆ received` always holds for an `Ok` proof, and the `remote.is_subset(&result.received)` check in `Capture::finish` can never fail.

The proof could just be the `remote` set, checked against `state.ids`. That removes a field, a check and one test branch (the `fetched` loop in `capture.rs`).

crates/pensieve-ingest/src/sync/worker/tests.rs
```diff
@@ -0,0 +1,438 @@
...
+    ledger.enqueue("fixture", relay, 100, 200).unwrap();
+    let lease = ledger.lease_next(now(), 600).unwrap().unwrap();
+    (ledger, lease)
+}
+
+fn uid() -> u32 {
+    UnixStream::pair().unwrap().0.peer_cred().unwrap().uid()
+}
+
+async fn send_inventory(stream: &mut UnixStream, lease: &super::super::jobs::Lease, relay: &str) {
```
erskingardner: Test gap: every relay test sends an empty inventory and the fixture relay stores at most one event. With negentropy 0.5 that finishes in a single round (client IdList, then server IdList, then client Skip, which returns `None`). So the parts of this loop that matter most are never run:

- the `Some(next)` NEG-MSG continuation
- a non-empty parent inventory where the sides overlap
- more than 128 missing IDs, which needs multiple fetch batches and `unsubscribe` between them

One fixture with a few hundred relay events and a partially overlapping inventory would cover all three.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-24T07:20:02Z` |
| Commit reviewed | `78b41103af8223aa50bff4b7bfd60639a0295e71` |

The corrected completion path holds up. I checked it against `nostr-relay-pool` 0.44.3 source. `handle_event_msg` awaits `save_event` before it broadcasts the `Event` message, and EOSE is handled in order after that. So every ID the fetch loop counts as received has already gone through `Capture`. Lag or closure, NEG-ERR, a status change, and an out-of-batch event all fail closed, and I did not find a path that sends a false `ProtocolDone`. Both of Grok's blockers are fixed. CI is green on this head.

The remaining problem is liveness. The worker fails safe, but some windows will fail on every retry:

1. **Events the SDK drops before our code sees them.** The relay can advertise an ID that the pinned SDK throws away: more than 2,000 tags (the live ingester has the same default limit, so the archive never has these events and they show up in every diff), a NIP-40 expiration in the past, or an event whose frame is over the 1 MiB IPC limit. Any of these makes the attempt `Incomplete` every time. The parent will keep splitting the window until it reaches `blocked`. The worker also returns the same error as it does for a transient failure, so the parent has no way to tell the two apart.
2. **The 120,000-character NEG-MSG limit rejects normal relay replies.** strfry replies with frames up to 500 KB and khatru up to 1 MiB. Large diffs will fail against both.

Neither causes a false completion. But this worker exists to close coverage gaps, and both problems hit the windows with the biggest gaps. Please decide how to handle them, and at least fix the relay limits, in this PR or explicitly in 3b before the scheduler starts using the worker.

Smaller notes are inline. `DownloadProof.received` is redundant, and the tests never exercise the multi-round path or more than one fetch batch.

Validation: I read the code and the pinned SDK, strfry and khatru sources. I did not run anything locally; CI passed on this exact head.

erskingardner:
Addressed the Claude review in 96d3f5f. Worker-specific SDK tag caps are removed while relay JSON is explicitly capped at 5 MiB and IPC remains 1 MiB. Incoming NEG-MSG replies no longer inherit our 60 KB outbound target. EOSE with missing advertised IDs now propagates Unavailable and executable exit 2, distinct from generic failure, never ProtocolDone. This cannot prove permanent absence or a volume limit. The pinned SDK still filters expired events; the execution plan explicitly gates scheduler activation on retaining these gaps, paced bounded retry/backoff and persistent-failure alerts, with no automatic split/skip. Proven-volume classification remains a 3b prerequisite. Added real-relay regressions for overlapping inventory/multiple diff rounds, multiple fetch batches, replies above 120000 hex characters, a 2001-tag event, and expired advertised events. All 12 focused tests and full just precommit passed immediately before commit. Retained the small redundant received-set proof check as defensive validation rather than taking the optional simplification. Claude exited 0 and was reaped; no reviewer process remains. Fresh CI and acceptance of the corrected head remain before merge. No production changes.

crates/pensieve-ingest/src/sync/worker.rs
```diff
@@ -0,0 +1,185 @@
...
+            .await
+            .map_err(|_| WorkerError::Incomplete)?;
+        tracing::info!(relay = %relay_url, phase = "sync", "isolated reconciliation");
+        let result = reconcile::download(&relay, &assignment, items).await;
+        tracing::info!(relay = %relay_url, phase = "disconnect", "isolated reconciliation");
+        client.disconnect().await;
+        // Close under the callback lock, not by hoping SDK Arc destruction closes
+        // its sender. Every callback accepted before this cut is already queued.
+        let summary = capture.finish(result.as_ref().ok()).await;
+        result?;
```
erskingardner: **Non-blocking: `Unavailable` also covers events that the worker itself rejected.**

When `Capture::capture` fails, `save_event` returns `Err`. In 0.44.3, `handle_event_msg` then returns early through `?` and sends no `Event` notification. The fetch loop never counts that ID, so EOSE produces `Unavailable`. Here `result?` takes precedence over the reason `Capture` already knows.

So exit 2 also covers:

- the 64 MiB `MAX_ATTEMPT_BYTES` budget running out. For example, a window of 30,000 kind-3 lists at about 3 KB each. The failure is deterministic and caused by volume, and splitting the window is the only fix.
- an out-of-window `created_at`, or the sticky failure after any earlier rejection.

The `WorkerError::Unavailable` doc and `negentropy_worker.md` both describe exit 2 as "EOSE arrived without all advertised IDs". Under the plan's "never split on exit 2" rule, a window that exceeds the byte budget can never finish. Meanwhile the analogous diff-count overflow (`remote.len() > MAX_EVENTS`) exits 1.

The cheapest fix is for `Capture` to keep its rejection reason, and for `attempt` to prefer it over `Unavailable`. A distinct budget error would give 3b the typed volume signal the plan already requires. If that is out of scope for 3a, at least list the attempt budget among the exit-2 causes in the doc.

crates/pensieve-ingest/src/sync/worker/reconcile.rs
```diff
@@ -0,0 +1,195 @@
...
+            RelayNotification::Message {
+                message:
+                    RelayMessage::NegMsg {
+                        subscription_id,
+                        message,
+                    },
+            } if subscription_id.as_ref() == &id => {
+                // Our outbound frame target does not constrain the responder.
+                // Accept larger bounded replies from strfry/khatru as well.
+                if message.len() > super::RELAY_MESSAGE_BYTES as usize {
```
erskingardner: Nit/simplification: this check can never fail. `handle_raw_relay_message` rejects the whole trimmed message above `limits.messages.max_size` before parsing, so the hex payload inside it can never be larger than 5 MiB.

`RELAY_MESSAGE_BYTES` is also exactly the SDK's default `MAX_MESSAGE_SIZE` (`5 * 1024 * 1024`). You could drop this check and the constant, and use:

```rust
RelayLimits { events: RelayEventLimits::disable(), ..Default::default() }
```

That states the actual intent (turn off event filters, keep the default wire cap) and removes about 8 lines.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-24T07:33:48Z` |
| Commit reviewed | `96d3f5ff946df635c37a915118b67c96ebfabe29` |

Posted as a comment because GitHub does not let the author's account approve its own PR. The recommendation is still merge.

This is a re-review of 96d3f5f, which responds to my review of 78b4110. The completion path I checked last time has not changed. The new commit fixes the two liveness problems I asked to be fixed here, and records the rest as 3b gates.

- **Tag cap:** `RelayLimits::disable()` removes the 2,000-tag and kind-3 caps and keeps the 5 MiB message cap. I checked this against `nostr-relay-pool` 0.44.3 `handle_event_msg`. The new 2,001-tag fixture event reaches `Capture`.
- **NEG-MSG size:** replies are now limited only by the message cap. The fixture has a reply over 120,000 hex characters, overlapping inventory with continuation rounds, and 19 fetch batches. That covers the test gap I raised.
- **Unfetchable IDs:** these now return `Unavailable` and exit 2, separate from the generic exit 1. The pinned SDK does not let the worker override the NIP-40 expiry drop, so documenting it as a gate for scheduler activation is a reasonable call for 3a.

There are two non-blocking notes inline. `Unavailable` also covers events the worker itself rejected, including when the 64 MiB attempt byte budget runs out. There is also a small cleanup that removes lines.

This PR accepts one limit that 3b needs to design around. A relay that keeps serving expired events (or any event the SDK drops) leaves its window unfinished for good. The worker does not report which IDs were missing, so the parent cannot record them as known exceptions. Coverage is the success metric, so the parent should get that information before activation, instead of only a paced retry and an alert.

Validation: I read the diff and the pinned SDK source. CI (rustfmt, clippy, doctests, 3 test shards) passed on this exact head. I did not run anything locally.

crates/pensieve-ingest/src/sync/worker/reconcile.rs
```diff
@@ -0,0 +1,195 @@
...
+                message:
+                    RelayMessage::NegErr {
+                        subscription_id, ..
+                    },
+            } if subscription_id.as_ref() == &id => return Err(WorkerError::Incomplete),
+            RelayNotification::RelayStatus { status } if status != RelayStatus::Connected => {
+                return Err(WorkerError::Incomplete);
+            }
+            RelayNotification::Shutdown => return Err(WorkerError::Incomplete),
+            _ => (),
```
erskingardner: Non-blocking. `RelayMessage::Closed` for this negentropy subscription falls through here, so a relay that rejects `NEG-OPEN` waits out the two-minute idle deadline. The fetch loop below already maps `Closed` on its subscription to `Incomplete`. The same match in this loop would fail closed immediately. This cannot become `ProtocolDone`; the idle deadline still bounds it.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-24T07:38:56Z` |
| Commit reviewed | `96d3f5ff946df635c37a915118b67c96ebfabe29` |
| Model | `Grok 4.7` |

The corrected completion path holds. In `nostr-relay-pool` 0.44.3, `handle_event_msg` awaits `save_event` before it broadcasts the event message, and this loop treats lagged or closed notification receivers as `Incomplete`. `ProtocolDone` is sent only after download proof, capture membership, and a matching ACK for every uploaded frame. `RelayLimits::disable()` drops the tag cap, the restored 5 MiB message cap covers strfry and khatru replies, and EOSE with a missing advertised ID exits 2 without a completion frame. Both earlier blockers stay fixed: duplicate inventory IDs are rejected at a later timestamp, and the SDK convenience sync loop is not used. Exact-head CI is green on all three test shards. One non-blocking note is inline.

GitHub rejected `APPROVE` because this account opened the pull request, so this review is submitted as a comment. The recommendation is still to merge.

erskingardner:
Self-review round complete on exact head 96d3f5ff946df635c37a915118b67c96ebfabe29. Claude Opus 5.5 and Grok 4.7 both recommend Merge; their reviews are comments because GitHub prevents author-account self-approval. Both CLI processes exited 0 and were reaped. All exact-head CI checks pass. No blocking findings and no source changes in this round. Carry the non-blocking items into the existing 3b activation gate: (1) preserve typed Capture rejection causes before allowing volume splitting; exit 2 currently also covers the 64 MiB attempt-budget exhaustion, out-of-window candidates and sticky callback rejection, not just SDK expiry or relay withholding; (2) report bounded missing-ID diagnostics to the parent, retaining unresolved obligations rather than silently treating exceptions as recovered; (3) fail immediately on a matching CLOSED during the diff instead of waiting for the already-bounded idle deadline. Keep the explicit defensive reply cap for now; its removal is optional cleanup. These approvals concern the inactive 3a worker slice, not deployment or scheduler activation.

PR #52, opened 2026-09-24
Preserve worker failure causes before scheduler integration

erskingardner:
## Scope
Bounded worker-outcome prerequisite for the isolated negentropy parent. No listener, scheduler, activation, database migration, or deployment.

## Changes
- Preserve verified capture failures through SDK notification suppression and EOSE.
- Exit 3 means distinct, verified, solicited event-byte exhaustion only. Count overflow stays generic failure; oversized individual events exit 4 and missing advertised events exit 2.
- Deduplicate repeated signed IDs before charging budgets and enable SDK subscription verification before capture.
- Handle matching CLOSED during reconciliation immediately.

## Validation
Full just precommit passed immediately before final commit 05d09039e854506d7c12cd738319c1736c51085c. Regression coverage includes duplicate floods, unsolicited and wrong-ID events through the real SDK, invalid candidates, budget classification, first-cause preservation, and CLOSED handling. Claude and Grok recommend merge on the final head; all GitHub CI checks pass.

## Activation gates
Authenticated process/attempt binding, durable diagnostics, parent scheduling and receipt-safe recovery, inventory gates, Linux isolation tests, and operator-approved canary remain separate work. No outcome proves archival or permits dropping gaps.


erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-24T07:51:56Z` |
| Commit reviewed | `b95bbbd919273ae9e6cf7380a59a9004dd64089f` |

_(I posted this as a comment because GitHub does not let you request changes on your own PR. Treat it as a request for changes.)_

Most of this is sound. The first-cause preservation works, EventSize is checked before the aggregate budgets, and the CLOSED-during-diff fix is small and has a real regression test. The blocker is that exit 3 is supposed to be the *proven volume* split signal, but a relay can forge it today:

1. **Duplicate frames count as volume.** `Capture::check_id` always returns `NotExistent`, so the pinned SDK (`nostr-relay-pool` 0.44.3 `handle_event_msg`) calls `save_event` again for every repeat of an event. `capture()` then increments `count`/`bytes` and forwards another frame, even though `state.ids.insert` already returned `false`. If a relay replays one valid ~900 KB in-window event about 70 times, the worker exits 3. Splitting can't fix that.
2. **Count exhaustion can't be reached by an honest relay.** `reconcile::download` already fails when `remote.len() > MAX_EVENTS`, and fetch batches partition `remote`. So distinct solicited captures are at most `MAX_EVENTS`, and the `state.count >= MAX_EVENTS` branch only fires on the (MAX_EVENTS+1)th frame, which must be a duplicate or unsolicited. Only byte exhaustion is real volume.

Suggested fix, which is also smaller: skip or reject events already in `state.ids` before touching the budgets, and leave count overflow as `Incomplete` so only byte exhaustion maps to `Volume`. Then update the docs and the `Volume` doc comment to say "byte budget".

Non-blocking: folding `finish_download` into `finish` (see inline) removes a function and the special-case match. It also makes the docs' "invalid candidates remain generic failures" claim true. Right now the PR's own test asserts that they come out as `Unavailable`/exit 2.

CI (clippy/test/doctests) was still pending at review time.

crates/pensieve-ingest/src/sync/worker/capture.rs
```diff
@@ -82,9 +98,19 @@ impl Capture {
...
+        .map_err(|error| match error {
+            ipc::ProtocolError::Limit => {
+                state.rejection = Some(Rejection::EventSize);
+                WorkerError::EventSize
+            }
+            other => WorkerError::Protocol(other),
         })?;
-        if state.bytes + wire.len() as u64 > ipc::MAX_ATTEMPT_BYTES {
-            return Err(WorkerError::Incomplete);
+        if state.count >= ipc::MAX_EVENTS
```
erskingardner: **Blocking.** This budget isn't a "verified in-window candidate" count. Because `check_id` always says `NotExistent`, the SDK re-invokes `save_event` for every duplicate delivery, and each one increments `count`/`bytes` here (line 135 ignores whether `ids.insert` returned `false`). A relay replaying one large valid event can therefore produce `Volume`/exit 3, the one outcome the docs call a split signal.

Also, `reconcile::download` already caps `remote` at `MAX_EVENTS` and only fetches those IDs. So `state.count >= MAX_EVENTS` can only fire on the (MAX_EVENTS+1)th frame, which has to be a duplicate or unsolicited event. That makes count exhaustion a relay-misbehavior signal, not volume.

Suggestion: reject or skip `state.ids.contains(&event.id)` before the budget checks, and keep count overflow as `Incomplete`. Then `Volume` means byte exhaustion only:

```rust
if state.count >= ipc::MAX_EVENTS || state.ids.contains(&event.id) {
    return Err(WorkerError::Incomplete);
}
// ... encode (EventSize) ...
if state.bytes + wire.len() as u64 > ipc::MAX_ATTEMPT_BYTES {
    state.rejection = Some(Rejection::Volume);
    return Err(WorkerError::Volume);
}
```

It would be worth a regression test that feeds the same valid event twice and asserts it never yields `Volume`.

crates/pensieve-ingest/src/sync/worker/capture.rs
```diff
@@ -128,6 +157,22 @@ impl Capture {
         }
         Ok((state.count, state.digest))
     }
+
+    pub async fn finish_download(
```
erskingardner: Non-blocking simplification. `finish` already returns the rejection before anything else, so this wrapper only exists to reorder precedence against `result?`. Production only calls `finish_download`. A single `finish` that takes the `Result` and puts *any* local failure first is shorter and removes the special-case match:

```rust
pub async fn finish(&self, result: Result<DownloadProof, WorkerError>) -> Result<(u64, [u8; 32]), WorkerError> {
    let mut state = self.state.lock().await;
    state.sender.take();
    if state.failed {
        return Err(state.rejection.map_or(WorkerError::Incomplete, Rejection::error));
    }
    let result = result?;
    if !result.remote.is_subset(&result.received) || !result.remote.is_subset(&state.ids) {
        return Err(WorkerError::Incomplete);
    }
    Ok((state.count, state.digest))
}
```

Tests would pass `Ok(DownloadProof::default())` instead of `Some(&...)`. This also fixes the doc mismatch below: a locally rejected candidate would exit 1 instead of being reported as relay unavailability (exit 2).

docs/negentropy_worker.md
```diff
@@ -63,9 +63,19 @@ qualify. Unsolicited frames, cancellation, EOF and wrong ACKs fail closed. Failu
...
-rejected by the IPC cap; it is neither proof of permanent absence nor proof of a
-volume limit. Other failures exit 1. Never split/skip/complete solely from exit 2.
+It does not distinguish relay withholding from SDK-expired NIP-40 events; it is
+neither proof of permanent absence nor proof of a volume limit. Never split/skip/
+complete solely from exit 2. Verified local count/byte exhaustion exits 3; a verified
+individual event exceeding the IPC cap exits 4. Capture preserves its first local
+rejection even if the SDK suppresses Event notification and EOSE later reports
+Unavailable. Signature/window checks precede classification; a single oversized
+event is checked before aggregate budgets because splitting cannot fix it.
+Cancellation/invalid candidates remain generic failures, never inferred volume.
```
erskingardner: This doesn't match the code. The PR's own test (`capture.rs`, the invalid-event case at the end of `local_rejections_survive_eose_and_keep_the_first_cause`) asserts that an invalid candidate followed by EOSE yields `Unavailable`, i.e. exit 2, not a generic exit-1 failure. Out-of-window candidates behave the same way. The edit above also removed "an event rejected by the IPC cap" from the list of exit-2 causes, which implies the remaining list (withholding, NIP-40 expiry) is complete, and it isn't. Either change the precedence (see the `finish` comment) or document that local non-volume rejections can still show up as exit 2.

erskingardner: This sentence says an invalid or out-of-window candidate stays a generic failure. `finish_download` only keeps `Volume` and `EventSize`; every other capture error is replaced by the download result. The new test locks that in: a signature failure followed by `Unavailable` stays `Unavailable`, so the process exits 2, not 1.

Say that directly. Otherwise the scheduler can treat this case as generic loss (exit 1) when the worker will actually report exit 2. It still must not be treated as a split signal.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-24T07:56:34Z` |
| Commit reviewed | `b95bbbd919273ae9e6cf7380a59a9004dd64089f` |
| Model | `Grok 4.7` |

Typed exits are the right bounded prerequisite before any split policy. This stays inside that boundary: volume and single-event size are distinguished, unverified relay diff overflow stays generic, and nothing here activates a scheduler.

Exit 3 and 4 survive the pinned SDK path that logs a rejected `save_event`, suppresses the event notification, and later surfaces EOSE as unavailable. Individual frame size is decided before the attempt budget, and a matching `CLOSED` during the diff fails immediately.

One documentation mismatch is noted inline. GitHub will not accept an approval from the pull request author, so this review is submitted as a comment. The recommendation is still to merge.

crates/pensieve-ingest/src/sync/worker/capture.rs
```diff
@@ -65,27 +82,45 @@ impl Capture {
...
+            other => WorkerError::Protocol(other),
         })?;
-        if state.bytes + wire.len() as u64 > ipc::MAX_ATTEMPT_BYTES {
+        // The diff is already capped at MAX_EVENTS. More distinct callbacks
+        // cannot prove an honest dense window; do not turn them into a split.
+        if state.count >= ipc::MAX_EVENTS {
             return Err(WorkerError::Incomplete);
         }
+        if state.bytes + wire.len() as u64 > ipc::MAX_ATTEMPT_BYTES {
+            state.rejection = Some(Rejection::Volume);
```
erskingardner: **Blocking.** `Volume` is described as "distinct verified in-window candidates". But nothing here checks that the event was actually requested. The worker adds the relay with default `RelayOptions`, and in `nostr-relay-pool` 0.44.3 `verify_subscriptions` defaults to `false`. So `handle_event_msg` saves every `EVENT` into `Capture`, whatever its subscription ID. That includes events sent during the NEG diff (before any fetch exists) and events on subscription IDs the worker never opened. `reconcile::download` drops those notifications in its `_ => ()` arms.

Concrete forgery: the relay answers the diff with 0 or 1 IDs, then sends 64 distinct self-signed ~1 MiB events with `created_at` inside the window on subscription `"x"`. They pass the window and signature checks, fill `MAX_ATTEMPT_BYTES`, and the worker exits 3. The relay can do this for any sub-window, so a split would never converge. There's also a race: if the unsolicited event arrives on the real fetch ID, the event that crosses the budget is rejected in `save_event`, so its notification is suppressed and the `binary_search` guard in reconcile never runs.

A likely minimal fix is to have the SDK reject unsolicited events before `save_event`. Add the relay with `RelayOptions::new().limits(relay_limits()).verify_subscriptions(true)` through `client.pool().add_relay(..)` (or equivalent) instead of `client.add_relay`. Events on the NEG ID or on unknown IDs then fail `SubscriptionNotFound`, and fetch events must match `Filter::ids(batch)`. Please add a regression to the relay fixture that sends unsolicited in-window events on a foreign subscription ID and asserts the exit code is never 3.

docs/negentropy_worker.md
```diff
@@ -63,9 +63,21 @@ qualify. Unsolicited frames, cancellation, EOF and wrong ACKs fail closed. Failu
...
+Repeated verified event IDs consume no additional count/byte budget. Count overflow
+is a generic failure: the verified diff already caps the number of solicited IDs.
+All other failures exit 1. Unknown exits/signals are not volume signals either.
+These exit codes require parent attempt/process binding before use in split policy;
+they do not replace durable receipt reconciliation. Relay-advertised diff-count
+overflow is not verified event volume and still fails generically. A matching
+CLOSED during the diff now fails immediately instead of waiting for idle expiry.
```
erskingardner: Non-blocking: this paragraph grew from 4 lines to 16, and parts of it say the same thing twice. "Never split/skip/complete solely from exit 2" and "Unknown exits/signals are not volume signals" cover the same ground, and there are two separate count-overflow sentences (lines 74-75 and 78-79). Phrases like "now fails immediately instead of waiting for idle expiry" are changelog wording that will go stale. Consider cutting it down to a table of exit codes (0/1/2/3/4 → meaning → may split?) plus the one sentence about binding before split policy. `negentropy_execution_plan.md` repeats the same content ("The worker now preserves..."); a link back here would be enough.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-24T08:07:33Z` |
| Commit reviewed | `cb5e15fc2e30c45d819b0007025ebe15a67458b1` |

_(Posted as a comment because GitHub does not allow requesting changes on your own PR. Please treat it as a request for changes.)_

cb5e15f fixes both issues from the previous round. Repeated IDs no longer consume budget, and count overflow now maps to `Incomplete`. `finish_download` now checks local `failed` before `result?`, so an invalid candidate followed by EOSE exits 1. The two earlier doc threads on `negentropy_worker.md` line 73 are stale and can be resolved. The CLOSED-during-diff handling and its test look right: `download` has no idle timer of its own, so the 2 s timeout really does prove the fix.

**Blocking: unsolicited events can still forge exit 3.** The worker builds its relay with default `RelayOptions`. In the pinned `nostr-relay-pool` 0.44.3, `verify_subscriptions` defaults to `false`, so `handle_event_msg` calls `Capture::save_event` for *any* `EVENT`, whatever its subscription ID. That includes events sent during the NEG diff and events on made-up subscription IDs. `reconcile::download` ignores those notifications (`_ => ()`). So a relay can advertise a near-empty diff, push about 64 distinct self-signed ~1 MiB in-window events on a junk subscription ID, and get `Volume`/exit 3. The window was never dense, and the relay can repeat this for every sub-window, so a split policy would never converge. The in-batch `binary_search` check doesn't help either: an unsolicited event that crosses the budget is rejected in `save_event`, so its notification is suppressed and reconcile never sees it. This is the same forgery class as the duplicate issue, just through a different path. Details and a likely one-line fix are inline.

Non-blocking: the new doc paragraphs are long, repeat each other, and read like a changelog (see inline).

CI: rustfmt, clippy and doctests pass. The three test shards were still pending at review time. I did not run anything locally.

crates/pensieve-ingest/src/sync/worker/capture.rs
```diff
@@ -65,27 +82,45 @@ impl Capture {
...
+            }
+            other => WorkerError::Protocol(other),
         })?;
-        if state.bytes + wire.len() as u64 > ipc::MAX_ATTEMPT_BYTES {
+        // The diff is already capped at MAX_EVENTS. More distinct callbacks
+        // cannot prove an honest dense window; do not turn them into a split.
+        if state.count >= ipc::MAX_EVENTS {
             return Err(WorkerError::Incomplete);
         }
+        if state.bytes + wire.len() as u64 > ipc::MAX_ATTEMPT_BYTES {
```
erskingardner: **Blocking.** A relay can still force this branch, and therefore exit 3, without a dense diff.

`ClientOptions::verify_subscriptions` defaults to false, and the worker never enables it. In nostr-relay-pool 0.44.3, `handle_event_msg` then calls `save_event` for every `EVENT` before any filter check. `reconcile::download`'s diff loop ignores those notifications (`_ => ()`). During `NEG-OPEN` a relay can push distinct, valid, in-window events that are absent from the reconciled id set, and each one is charged here.

The 8 MiB in-flight cap does not cap this counter. Parent ACKs release credit, `state.bytes` climbs to `MAX_ATTEMPT_BYTES`, and this branch records `Volume`. `finish_download` returns that rejection before the download result, so the diff's own failure never replaces it. The comment directly above already says an extra callback cannot prove an honest dense window. These bytes cannot either. Splitting the window will not stop a relay that injects a fresh batch on every attempt.

Smallest fix: `.verify_subscriptions(true)`, so an `EVENT` with no subscription is dropped before `save_event`, and the fetch filter (the solicited ids) is enforced. Add a regression where distinct in-window `EVENT`s during `NEG-OPEN` stay exit 1.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-24T08:13:47Z` |
| Commit reviewed | `cb5e15fc2e30c45d819b0007025ebe15a67458b1` |
| Model | `Grok 4.7` |

_(GitHub will not accept a request-changes review from the pull request author. Treat this as a request for changes.)_

The duplicate-replay and count-exhaustion holes from the earlier review are closed. Exit 3 can still be forced by distinct events that were never in the reconciled diff: subscription filtering is off, and the diff loop ignores `EVENT`s. A scheduler cannot treat that exit as a split signal until those bytes are limited to solicited ids.

`CLOSED` during the diff, first-cause preservation, and invalid candidates staying on exit 1 match the in-repo docs. The PR description still says count exhaustion is exit 3; the code and `docs/negentropy_worker.md` correctly limit exit 3 to byte exhaustion.

CI was still pending for clippy, tests, and doctests. rustfmt had passed.

crates/pensieve-ingest/src/sync/worker/capture.rs
erskingardner: Resolved in 05d0903. In `nostr-relay-pool` 0.44.3, `ClientOptions::verify_subscriptions(true)` reaches the relay through `compose_relay_opts`. `handle_event_msg` then rejects `EVENT`s whose subscription ID is not registered (`SubscriptionNotFound`) before `save_event` runs. That covers the raw `NEG-OPEN` ID and junk IDs. It also rejects events that fail the fetch `Filter::ids(batch)` (`EventNotMatchFilter`). `unsubscribe` removes finished fetch IDs, so late replays of an earlier batch are rejected too. The SDK logs the error and drops the event without disconnecting, and the worker has SDK log targets turned off. The new fixture injects a distinct in-window event on both paths. It asserts `count == 1` or `received == 0`, and would fail without the flag.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-24T08:22:05Z` |
| Commit reviewed | `05d09039e854506d7c12cd738319c1736c51085c` |

_(GitHub does not allow approving your own PR, so this is posted as a comment. The recommendation is still to merge.)_

05d0903 fixes the last blocker, which was unsolicited events forging exit 3. It does this with one SDK option instead of new code, and that is the right size of fix. I checked it against the pinned `nostr-relay-pool` 0.44.3 source:

- `ClientOptions::verify_subscriptions(true)` is copied into every relay the client adds (`compose_relay_opts`).
- `handle_event_msg` rejects an event before `save_event` when its subscription ID is not registered. That covers the raw `NEG-OPEN` ID and made-up IDs. It also rejects events that don't match the fetch's `Filter::ids(batch)`. After EOSE, `unsubscribe` removes the fetch ID, so replaying an earlier batch is rejected too.
- A rejected event is logged and dropped. The connection stays up, and the worker's SDK log targets are `off`, so the payloads don't leak.
- The new fixture injects a distinct in-window event both on a junk ID and on the real NEG/fetch ID. It asserts `count == 1` (complete path) and `received == 0` (partial path), so it fails without the flag.

With this change, `Volume` can only come from bytes of distinct, signed, in-window events whose IDs the relay advertised in the diff. That is the strongest signal the worker can give. Turning it into split policy is correctly left to the parent-binding work.

Non-blocking:
- The PR description still says "attempt count/byte exhaustion (exit 3)". Count overflow is now exit 1, so the description should say byte exhaustion only.
- The earlier doc-length comment on `negentropy_worker.md` is still open, and this commit made the paragraph 4 lines longer. A small exit-code table would still read better than the running prose. Also, "now fails immediately instead of waiting for idle expiry" reads like changelog text.

CI: rustfmt, clippy and doctests pass. The three test shards were still pending at review time. I did not run anything locally.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-24T08:24:48Z` |
| Commit reviewed | `05d09039e854506d7c12cd738319c1736c51085c` |
| Model | `Grok 4.7` |

_(GitHub will not accept an approval from the pull request author. Treat this as an approval.)_

`05d09039` closes the remaining exit-3 forgery. In nostr-relay-pool 0.44.3, `handle_event_msg` checks `verify_subscriptions` before `save_event`, and `ClientOptions::verify_subscriptions` is copied onto the relay at add time. `NEG-OPEN` is sent with `send_msg` and never enters the subscription map, so diff-phase `EVENT`s fail `SubscriptionNotFound` and never reach capture. Fetch `REQ`s use an ids-only filter, and `ids_match` requires `event.id` to be in that set. Duplicates still skip the budget, and count overflow stays `Incomplete`. Exit 3 is distinct, verified, solicited byte exhaustion.

The updated SDK test is the right proof: the fixture pushes a valid in-window event on a junk subscription and a wrong id on the live fetch subscription, the success path asserts a single admitted frame, and the partial path asserts `received == 0`.

Non-blocking: the PR description still says count exhaustion is exit 3. The code and `docs/negentropy_worker.md` correctly keep that on exit 1. rustfmt has passed; clippy, tests, and doctests were still pending. I did not run anything locally.

PR #53, opened 2026-09-24
Retain bounded worker failure diagnostics across retries

erskingardner:
## Scope
Bounded terminal failure diagnostics and durable attempt fencing; no listener, scheduler, production activation, or deployment.

## Changes
A single AttemptFailed report records typed outcomes and bounded sorted missing-ID hints before fencing the attempt terminal. Receipt obligations survive retry and reopen, including failure with zero receipts. Worker failures drain already queued candidates through acknowledgements before reporting where the connection and deadline permit.

Schema 4 is fresh-only: older prototype ledgers fail closed without mutation, not automatic migration. Existing schema-4 ledgers may reopen above admission limits so expiry and bounded recovery remain available.

## Validation
Full just precommit passed after final review fixes and master integration. Tests cover terminal reconnect fencing, schema rejection preservation, over-ceiling recovery, wrong sequence and lease, rollback and budget limits, and real-SDK partial fetch behavior. Missing-ID hints never authorize skipping a gap. Independent final-head reviews and green CI required before merge.


crates/pensieve-ingest/src/sync/jobs.rs
```diff
@@ -231,10 +238,17 @@ impl JobLedger {
                     );
                     INSERT INTO receipt_totals VALUES(1,0);")?;
                     tx.pragma_update(None, "application_id", APPLICATION_ID)?;
-                    tx.pragma_update(None, "user_version", 3)?;
+                    tx.execute_batch(FAILURE_SCHEMA)?;
+                    tx.pragma_update(None, "user_version", 4)?;
+                    check_budget(&tx, path, limits, 0)?;
+                }
+                (APPLICATION_ID, 3) => {
```
erskingardner: **Blocking:** this migration goes against the policy written in `docs/negentropy_archive_receipts.md`: "The undeployed prototype schema is version 3. Versions 1 and 2 fail closed without mutation; there is intentionally no speculative migration for unshipped databases." No v3 ledger exists in production (this PR says nothing is activated). Suggest deleting this arm, so `(APPLICATION_ID, 3)` falls through to the unsupported-version error like v1/v2, and deleting `v3_migration_preserves_unfinished_jobs_and_receipts`. The create-path bump to 4 is enough. Also update that doc sentence, which still says version 3.

crates/pensieve-ingest/src/sync/ipc.rs
```diff
@@ -111,14 +117,20 @@ pub enum Message {
     },
     /// Incomplete attempt. Never interpreted as successful empty output.
     AttemptFailed { header: Header, reason: Failure },
+    /// Bounded diagnostic accounting, not archival proof or a retry decision.
+    AttemptReport {
```
erskingardner: **Blocking:** this adds a second terminal failure message next to `AttemptFailed` (line 119) with different guarantees. `AttemptReport` is fenced in the ledger and requires `outstanding` to be empty. `AttemptFailed` records nothing, skips the outstanding check, and still returns the same `UploadAction::Failed(_)`. A caller can't tell whether the attempt is now terminal, which undercuts the "a report is terminal for that attempt" guarantee in the doc. Nothing in the worker sends `AttemptFailed` any more (only tests do), and the wire format is not deployed. Suggest changing it to `AttemptFailed { header, report: FailureDiagnostic }`, with a single handler that does what the `AttemptReport` arm does now.

crates/pensieve-ingest/src/sync/ipc.rs
```diff
@@ -380,6 +392,21 @@ impl UploadSession {
...
+            Message::AttemptReport { header, report } => {
+                if !self.outstanding.is_empty() {
+                    return Err(ProtocolError::State);
+                }
+                ledger.record_failure_report(&self.lease, header.sequence, &report, now)?;
+                use super::failure::FailureKind;
+                let reason = match report.kind {
+                    FailureKind::Relay => Failure::Relay,
+                    FailureKind::Volume => Failure::Volume,
+                    FailureKind::EventSize => Failure::EventSize,
```
erskingardner: This mapping, and the three new `Failure` variants (`Unavailable`/`Volume`/`EventSize`), exist only to convert `FailureKind` back into a parallel enum. Once there is a single terminal message, `UploadAction::Failed` can carry `FailureKind` (or the whole report) and `Failure` can go away or shrink.

crates/pensieve-ingest/src/sync/worker.rs
```diff
@@ -51,6 +52,9 @@ pub enum WorkerError {
     /// identical. Preserve the gap; do not automatically split or skip it.
     #[error("worker advertised events unavailable at EOSE")]
     Unavailable,
+    /// Bounded outstanding-ID diagnostic at EOSE, never logged as payload.
+    #[error("worker advertised events remain outstanding at EOSE")]
+    Outstanding(FailureDiagnostic),
```
erskingardner: `WorkerError::Unavailable` (line 54) is now only constructed in `capture.rs` tests. Production EOSE-missing now always produces `Outstanding`. Having two variants for the same condition means maintaining two match arms (`exit_code`, `diagnostic`) and a zero-sample `Unavailable` report that production never creates. Suggest folding them: `Unavailable(FailureDiagnostic)`.

crates/pensieve-ingest/src/sync/worker/reconcile.rs
```diff
@@ -151,7 +151,17 @@ pub(super) async fn download(
...
                     if !batch.iter().all(|id| received.contains(id)) {
-                        return Err(WorkerError::Unavailable);
+                        let remaining: Vec<_> =
+                            ids.iter().filter(|id| !received.contains(id)).collect();
+                        return Err(WorkerError::Outstanding(super::FailureDiagnostic {
+                            kind: super::FailureKind::Unavailable,
+                            missing_count: remaining.len() as u64,
+                            sample: remaining
+                                .into_iter()
+                                .take(128)
```
erskingardner: Use `MAX_FAILURE_SAMPLE` here instead of the literal `128`. Likewise `MAX_FAILURE_MISSING` (failure.rs:8) should just be `ipc::MAX_EVENTS`, which is the value that actually bounds `remote`. If either copy drifts, the parent's `validate()` rejects a legitimate report and the diagnostic turns into a protocol error.

crates/pensieve-ingest/src/sync/jobs.rs
```diff
@@ -427,6 +441,62 @@ impl JobLedger {
...
+    }
+
+    /// Read retained diagnostic history, including failed attempts after retries.
+    pub fn failure_report(
+        &self,
+        job: i64,
+        attempt: i64,
+    ) -> Result<Option<FailureDiagnostic>, LedgerError> {
+        let json: Option<String> = self.db.query_row(
+            "SELECT CASE WHEN length(report)<=32768 THEN report ELSE NULL END FROM failure_reports WHERE job=?1 AND attempt=?2",
```
erskingardner: Nit: the size bound is enforced three times: `MAX_FAILURE_JSON`, `CHECK(length(report)<=32768)`, and this `CASE WHEN`. With `validate()` capping the sample at 128 IDs, the JSON stays around 17 KB, so none of them can trigger. If a row ever did exceed the bound, the `NULL` here would come back as a column-type error, not a clean `None`. Keep the `CHECK` as the schema backstop and drop the other two. Storing IDs as hex would also halve the row size compared with decimal byte arrays.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-24T08:13:00Z` |
| Commit reviewed | `4ad4d35bd4af84e6cfd0443a0072c3cba9efed45` |

The core idea is sound and the fencing works: one bounded, validated diagnostic is recorded in the same ledger transaction as receipts, it is terminal for the attempt, and it never counts as completion. Wrapping the SDK lifecycle in an inner `async` block also fixes a real bug. Before, an `add_relay`/`try_connect` error skipped `finish_download`, so the events channel never closed and the worker waited out the idle deadline.

The change adds parallel paths next to the ones it replaces instead of changing them, and it adds a migration the project explicitly says it does not want.

**Blocking**
1. **Speculative v3 to v4 migration** (`jobs.rs`). `docs/negentropy_archive_receipts.md` says: "The undeployed prototype schema is version 3 ... there is intentionally no speculative migration for unshipped databases." v1 and v2 are rejected on that basis, and this PR's description confirms nothing is activated. Drop the `(APPLICATION_ID, 3)` arm and `v3_migration_preserves_unfinished_jobs_and_receipts`, keep only the create-path bump to 4, and update the stale "version 3" sentence in that doc.
2. **Two terminal failure messages with different guarantees** (`ipc.rs`). `AttemptFailed` is still accepted but is not fenced. It records nothing, skips the empty-`outstanding` check, and returns the same `UploadAction::Failed(_)` as the fenced `AttemptReport`. So a future scheduler that sees `Failed` can't tell whether the attempt is terminal in the ledger. The worker never sends `AttemptFailed`, so replace its payload with the report (`AttemptFailed { header, report }`), keep the single fenced path, and delete the `FailureKind` to `Failure` mapping and the three new `Failure` variants.

**Non-blocking**
- `WorkerError::Unavailable` is now only built in tests. `Outstanding` replaced it in production, so fold them into one variant.
- `take(128)` and `MAX_FAILURE_MISSING = 50_000` duplicate `MAX_FAILURE_SAMPLE` and `ipc::MAX_EVENTS`. If either drifts, a valid report fails `validate()` and becomes a protocol error.
- The size bound is enforced three times (`MAX_FAILURE_JSON`, the SQL `CHECK`, and `CASE WHEN length(...)` on read), and none of the three can trigger once `validate()` passes.

Validation: clippy, rustfmt and doctests pass on this SHA. The three test shards were still pending when this was posted. No local execution was done.

crates/pensieve-ingest/src/sync/ipc.rs
```diff
@@ -380,6 +392,21 @@ impl UploadSession {
                 ledger.verify_active(&self.lease, now)?;
                 Ok(UploadAction::Failed(reason))
             }
+            Message::AttemptReport { header, report } => {
```
erskingardner: `AttemptReport` inserts a terminal row, so a later `ProtocolDone` on this lease fails `has_failure`. The `AttemptFailed` arm above still returns `Failed` after `verify_active` and leaves the attempt open.

That is a false completion when nothing has been received. A new `UploadSession` starts at sequence 1, and `record_protocol_done` accepts an empty digest when `received` is 0. The same lease can move to `awaiting_durability` after the parent has already observed `Failed`. Reconnect is specified as expire-and-retry; this row is the backstop when that step is skipped. With receipts already stored, the fresh session count also blocks resume, so the existing reconnect test (sequence 1 after one receipt) does not isolate the fence.

Make `AttemptFailed` terminal as well: reject the variant, or insert a failure row under the same lease and sequence checks before returning `Failed`. Add a zero-receipt case that opens a new session, sends empty `ProtocolDone`, and expects rejection.

crates/pensieve-ingest/src/sync/jobs.rs
```diff
@@ -231,10 +238,17 @@ impl JobLedger {
                     );
                     INSERT INTO receipt_totals VALUES(1,0);")?;
                     tx.pragma_update(None, "application_id", APPLICATION_ID)?;
-                    tx.pragma_update(None, "user_version", 3)?;
+                    tx.execute_batch(FAILURE_SCHEMA)?;
+                    tx.pragma_update(None, "user_version", 4)?;
+                    check_budget(&tx, path, limits, 0)?;
+                }
+                (APPLICATION_ID, 3) => {
+                    check_budget(&tx, path, limits, 0)?;
```
erskingardner: Both `check_budget` calls sit inside `open`'s migration transaction. A budget error rolls the transaction back, so `user_version` stays 3. This binary opens version 3 only by migrating it, so a ledger already at the admission ceiling cannot be opened and `expire` / `retry` never run.

Those recovery paths skip the ceiling on purpose: a full ledger must still be able to release its lease. The byte comment on `LedgerLimits` says the same thing — the ceiling is an admission limit, and SQLite can grow past it during commit. Commit the empty `failure_reports` table on the 3→4 upgrade even when a later admission would be refused. Keep the ceiling on `record_failure_report`.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-24T08:17:37Z` |
| Commit reviewed | `4ad4d35bd4af84e6cfd0443a0072c3cba9efed45` |
| Model | `Grok 4.7` |

Durable `AttemptReport` rows are the right slice: bounded classes, lease and sequence checks, receipts kept, and protocol completion blocked after a stored report. Two parent paths still break that invariant.

- `AttemptFailed` is accepted as terminal and writes no fence. With zero receipts, a new `UploadSession` can `ProtocolDone` an empty digest and move the job to `awaiting_durability`.
- Schema 3 opens only by migrating, and that migration is inside the admission budget check. A ledger already at the ceiling cannot be opened, so `expire` / `retry` never run.

Clippy, rustfmt, and doctests passed on this SHA. The three test shards were still pending.

crates/pensieve-ingest/src/sync/worker.rs
```diff
@@ -178,15 +197,36 @@ async fn attempt(
...
+                write_message(
+                    &mut stream,
+                    &Message::AttemptFailed {
+                        header: assignment.header(admitted + 1),
+                        report: error.diagnostic(),
+                    },
+                ),
+            )
+            .await
+            .map_err(|_| WorkerError::Deadline)??;
```
erskingardner: If this write fails or hits the idle deadline, `??` returns `Protocol`/`Deadline` instead of `error`. The worker binary then exits 1 instead of 2/3/4, so the specific cause is lost in exactly the case where the report can't be delivered either. The doc already calls the report best-effort, so the exit code shouldn't depend on it:

```rust
let _ = timeout_at(idle, write_message(&mut stream, &Message::AttemptFailed { .. })).await;
return Err(error);
```

crates/pensieve-ingest/src/sync/worker/tests.rs
```diff
@@ -467,7 +482,7 @@ async fn real_sdk_rejects_unsolicited_events_and_requires_archive_admission() {
         server.abort();
         let _ = server.await;
         if partial {
-            assert!(matches!(outcome, Err(WorkerError::Unavailable)));
+            assert!(matches!(outcome, Err(WorkerError::Unavailable(_))));
```
erskingardner: This doesn't prove the report arrived. The parent loop treats `Err(_) if partial => break` (L422) as a normal exit, so if the worker stopped sending `AttemptFailed` and just closed the socket, this test would still pass. Please assert `ledger.failure_report(id, 1).unwrap().is_some()` here.

Also, the partial relay delivers zero events, so end-to-end the report sequence is only ever `admitted + 1 == 1`. A case with one delivered event and one withheld would cover the drain-then-report sequencing this PR is about.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-24T08:33:55Z` |
| Commit reviewed | `671f761e74f4e751e0253525d73c182fac34c2f5` |

Both blocking items from the previous round are fixed. There is now one fenced `AttemptFailed` path, schema 4 is fresh-only with no v3 migration, and over-ceiling ledgers reopen for `expire`/`retry`. The ledger side is solid. Once a report is stored, further receipts, protocol completion and repeat reports are all rejected. The zero-receipt reopen test closes the empty-`ProtocolDone` hole. The premise is reasonable for a staged slice: the fence has real value even before a scheduler reads `failure_report`.

**Blocking (both small)**
1. `worker.rs`: a failed report write replaces the original error, so the process exit code (2/3/4) is lost.
2. `worker/tests.rs`: no test proves the worker actually delivers `AttemptFailed`. The real-SDK partial test also passes on a plain EOF, and nothing covers a failure after one or more admitted events (the `admitted + 1` sequencing).

**Non-blocking**
- `FailureKind::Cancelled` is never constructed.
- The JSON size bound is enforced three times, and none of them can trigger once `validate()` passes.
- The last section of the new doc describes the PR, not the design.
- `WorkerError::Unavailable(FailureDiagnostic)` repeats `kind`, which forces the same 5-line literal into five test sites in `capture.rs`/`tests.rs`. A small test helper would remove most of that churn.

Validation: rustfmt passes on this SHA. Clippy, doctests and the three test shards were still pending when this was posted. No local execution was done. This is posted as a comment because GitHub does not allow request-changes on your own PR.

crates/pensieve-ingest/src/sync/failure.rs
```diff
@@ -0,0 +1,87 @@
...
+    /// Relay or reconciliation lifecycle failure.
+    Relay,
+    /// Attempt volume exceeded a configured bound.
+    Volume,
+    /// An event exceeded the permitted frame size.
+    EventSize,
+    /// Advertised event IDs remain outstanding; not proof they are undeliverable.
+    Unavailable,
+    /// The worker was cancelled.
+    Cancelled,
```
erskingardner: `Cancelled` is never constructed. `WorkerError::diagnostic()` maps every other error to `Relay`, and nothing on the parent side builds a report. Drop it until something emits it.

crates/pensieve-ingest/src/sync/jobs.rs
```diff
@@ -427,6 +435,62 @@ impl JobLedger {
...
+    }
+
+    /// Read retained diagnostic history, including failed attempts after retries.
+    pub fn failure_report(
+        &self,
+        job: i64,
+        attempt: i64,
+    ) -> Result<Option<FailureDiagnostic>, LedgerError> {
+        let json: Option<String> = self.db.query_row(
+            "SELECT CASE WHEN length(report)<=32768 THEN report ELSE NULL END FROM failure_reports WHERE job=?1 AND attempt=?2",
```
erskingardner: `validate()` caps a report at 128 IDs, which is roughly 17 KB of JSON. So `MAX_FAILURE_JSON` (L452), the `CHECK(length(report)<=32768)`, and this `CASE WHEN` can never trigger. If this one did, the `NULL` would come back as a confusing column-type error from `r.get::<String>` anyway. Keep `validate()` and at most the SQL `CHECK`, and drop the other two. This was also raised on the previous head.

docs/negentropy_failure_diagnostics.md
```diff
@@ -0,0 +1,43 @@
...
+undeployed prototype: earlier schema versions are rejected without mutation, not
+migrated automatically. Existing schema-4 ledgers open even when over the admission
+ceiling so lease expiry, retry and receipt recovery remain available.
+Diagnostic history survives retries and reopen; no retention deletion is added.
+The existing total ledger budget bounds growth and rejects new writes when full.
+The new wire variant fails closed on an older parent, so a future deployment must
+install matching parent/worker binaries before activation. This is not an online
+mixed-version rollout protocol.
+
+There is one terminal failure message, `AttemptFailed`, and every accepted failure
```
erskingardner: Nit: L36-43 describe the PR, not the design ("no legacy unfenced failure path" and the list of tests). They will go stale as soon as the code moves. Consider cutting them.

crates/pensieve-ingest/src/sync/worker/tests.rs
erskingardner: Non-blocking: this partial post-condition still passes when no report is stored. The loop above treats `read_wire` failure as success (`Err(_) if partial => break`), which was the old path where the worker exited `Unavailable` and closed the socket without a frame. These asserts only require the exit variant, `Leased`, and `received == 0`.

The new `Failed` arm is the only place that checks `failure_report`. Drop the read-error break so a missing `AttemptFailed` fails the test, and assert the stored report here (`missing_count == 1`, one sample ID) so the integration path cannot go green without the fence.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-24T08:40:09Z` |
| Commit reviewed | `671f761e74f4e751e0253525d73c182fac34c2f5` |
| Model | `Grok 4.7` |

The earlier blocking issues are fixed on this head. `AttemptFailed` is the only terminal failure message, and accepting it writes the attempt fence under the same lease, sequence, and budget checks as receipts. Schema 4 is create-only: version 3 is rejected without mutation, and an existing schema-4 ledger still opens above the admission ceiling so expiry and retry keep working. A failed fetch batch is processed in sorted ID order, so the IDs missing from that batch are a prefix of the 128-ID sample. Clippy, rustfmt, doctests, and all three test shards passed on this SHA.

GitHub will not accept an approval from the pull request author, so this is submitted as a comment. The recommendation is still to merge.

One non-blocking test gap is inline: the partial SDK test can still pass when the worker never sends the report.

crates/pensieve-ingest/src/sync/worker/capture.rs
```diff
@@ -263,7 +263,13 @@ mod tests {
             );
             assert_eq!(
                 capture
-                    .finish_download(Err(WorkerError::Unavailable))
+                    .finish_download(Err(WorkerError::Unavailable(
```
erskingardner: Nit: this 5-line `WorkerError::Unavailable(FailureDiagnostic { kind: Unavailable, missing_count: 0, sample: Vec::new() })` literal is repeated 4 times in this file and once more in `worker/tests.rs` (`expired_advertised_event_is_unavailable_not_success`). A `#[cfg(test)] fn unavailable() -> WorkerError` helper in `worker.rs` would cut about 30 lines and remove the fully qualified `crate::sync::failure::` paths.

crates/pensieve-ingest/src/sync/ipc.rs
```diff
@@ -527,6 +522,88 @@ mod tests {
...
+            !db.attempt_progress(lease.job().id, 1)
+                .unwrap()
+                .protocol_done
+        );
+        assert_eq!(db.get(lease.job().id).unwrap().state, JobState::Leased);
+    }
+
+    #[test]
+    fn diagnostic_failure_is_terminal_durable_and_never_completion() {
+        use super::super::failure::{FailureDiagnostic, FailureKind};
```
erskingardner: Nit: this import is redundant. The tests module already has `use super::*;`, which brings `FailureDiagnostic` and `FailureKind` in from the top-level import (the zero-receipt test above uses them unqualified).

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-24T08:53:17Z` |
| Commit reviewed | `7e59ab46de2a81f431c50b00a00330196c96a006` |

Both blocking items from the `671f761` round are fixed on this head.

- `report_failure` throws away the write result and returns the original `WorkerError`, so exit codes 2/3/4 survive a broken parent socket. `failed_diagnostic_delivery_preserves_the_original_exit_cause` covers this.
- The real-SDK test no longer treats a read error as a normal exit. It asserts the stored report (`missing_count == 1`, sample equal to the withheld ID), and the new `(partial, delivered) = (true, 1)` case covers the drain-then-report `admitted + 1` sequencing end to end.

I re-checked the sequencing. Each queued event is written and ACKed before the next `recv`, and `finish_download` drops the sender only after the SDK future finishes. So when the loop exits, `admitted` equals the parent's `received` and `outstanding` is empty. The sample is built from the sorted `ids` vector, so it passes `validate()`'s strict ordering check. Fresh-session sequencing (`count + 1`) together with the ledger's `received + 1` check keeps a reconnected session from recording a report after receipts exist.

**Still open from earlier rounds (non-blocking, all deletions):**
- `FailureKind::Cancelled` is still never constructed.
- The JSON size bound is still enforced three times, and none of the three can trigger once `validate()` passes.
- `docs/negentropy_failure_diagnostics.md` L36-43 still describe the PR rather than the design.
- The 5-line `Unavailable(FailureDiagnostic { .. })` literal now appears 6 times in the tests (see inline).

Validation: clippy, rustfmt and doctests pass on this SHA. The three test shards were still pending when this was posted, so please wait for them to go green before merging. No local execution was done.

This is posted as a comment because GitHub does not allow approving your own PR. The recommendation is still to merge.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-24T08:58:48Z` |
| Commit reviewed | `7e59ab46de2a81f431c50b00a00330196c96a006` |
| Model | `Grok 4.7` |

The failure fence on this head does the job this slice is for. `AttemptFailed` is the only terminal failure message, and accepting it stores the diagnostic under the same lease, sequence, and budget checks as receipts. A stored report rejects later receipts, protocol completion, and a second report. The zero-receipt reopen can no longer turn an empty `ProtocolDone` into `awaiting_durability`.

Schema 4 is created fresh. Older versions fail closed without writes. An existing schema-4 ledger still opens above the admission ceiling, so expiry and retry keep working.

The worker drains already admitted candidates, then sends the report. `report_failure` treats that write as best-effort and returns the original `WorkerError`, so a dead parent socket keeps exit codes 2, 3, and 4. The partial relay test now fails closed unless `AttemptFailed` arrives, including after one admitted event, and checks the stored sample against the withheld id.

Rustfmt, clippy, doctests, and all three test shards passed on this SHA.

Earlier non-blocking notes still stand. `FailureKind::Cancelled` has no constructor. The JSON size bound is checked in `validate`, `MAX_FAILURE_JSON`, the SQL `CHECK`, and the read query. The last section of `docs/negentropy_failure_diagnostics.md` describes this change and will go stale.

GitHub will not accept an approval from the pull request author, so this is submitted as a comment. The recommendation is still to merge.

PR #54, opened 2026-09-24
Add authenticated single-owner parent IPC sessions

erskingardner:
Stacked on #53 (durable failure diagnostics), with current master integrated. Inactive library-only parent executor: one blocking owner, capacity-one queue, exact writer/dedupe authority, peer UID before capability export, complete bounded inventory, archive-admission ACKs, deadlines/cancellation and joined shutdown. The single AttemptFailed report persists its schema-4 terminal fence and returns FailureKind. On its own deadline, serve cancels socket activity but awaits the owner result, preserving an already-committed terminal outcome. External future-drop still requires durable-state reread. Local recovery/invalid-request errors are typed separately from worker protocol errors. Tests include deterministic committed-ProtocolDone/deadline racing, real-worker parsing of 300 inventory IDs across multiple chunks, and buffered terminal frames after peer exit. All 11 focused parent tests and full just precommit passed on the updated code. No listener, scheduler, process launch, production migration or activation. Corrected-head CI and independent review acceptance remain required before merge; no reviewers launched by this update. Receipt maintenance, process binding and runtime readiness remain separate gates.


crates/pensieve-ingest/src/sync/parent.rs
```diff
@@ -0,0 +1,838 @@
...
+        if self.cancelled.load(Ordering::Acquire) || wall.is_none() {
+            return Err(std::io::Error::new(
+                std::io::ErrorKind::ConnectionAborted,
+                "parent session cancelled or expired",
+            ));
+        }
+        self.until
+            .checked_duration_since(Instant::now())
+            .filter(|d| !d.is_zero())
+            .map(|d| d.min(Duration::from_secs(wall.unwrap() as u64)))
```
erskingardner: Nit (dead code): every caller of `check()` discards the returned `Duration` (`self.check()?` / `socket.check()?`). That makes the `min(...)` against wall-clock remaining, and the `wall.unwrap()`, dead. Return `std::io::Result<()>` instead and just test `Instant::now() < self.until`. The code gets shorter and the `unwrap` goes away.

crates/pensieve-ingest/src/sync/parent.rs
```diff
@@ -0,0 +1,838 @@
...
+        ArchivedWindow::TooDense => return Err(ParentError::TooDense),
+    };
+    socket.check()?;
+    ledger.verify_active(&lease, now())?;
+    if writer.recovery_required() {
+        return Err(ParentError::Protocol(ProtocolError::State));
+    }
+    let assignment = Assignment::for_lease(&lease);
+    socket.send(&ParentMessage::Job {
+        assignment: Assignment::for_lease(&lease),
```
erskingardner: Nit: `assignment` is already built on the line above. Move it into the message (`ParentMessage::Job { assignment }`) and derive chunk headers from `Header::for_lease(&lease, sequence)`, or build it once and keep the local. Either way, drop the second `Assignment::for_lease` call.

crates/pensieve-ingest/src/sync/parent.rs
```diff
@@ -0,0 +1,838 @@
...
+    };
+    socket.check()?;
+    ledger.verify_active(&lease, now())?;
+    if lease.job().since < 0
+        || lease.job().until < lease.job().since
+        || lease.job().until - lease.job().since >= 900
+    {
+        return Err(ParentError::Protocol(ProtocolError::State));
+    }
+    if writer.recovery_required() {
```
erskingardner: Non-blocking, but it matters for the next slice. `writer.recovery_required()` (here and at L405), invalid job window bounds (L372-376) and a bad caller `timeout` (L211-213) all come back as `ParentError::Protocol(ProtocolError::State)`, which is the same value a misbehaving worker produces. #52 went out of its way to keep failure causes typed for scheduler decisions. A scheduler that gets this error can't tell "pause admissions, archive needs recovery" from "caller bug" from "worker violated protocol". A `ParentError::RecoveryRequired` variant, plus an `InvalidRequest` (or reusing `Archive(Config)`) for caller-side bounds, would keep those apart at almost no cost.

crates/pensieve-ingest/src/sync/parent.rs
```diff
@@ -0,0 +1,838 @@
...
+    pub async fn shutdown(mut self) -> Result<JobLedger, ParentError> {
+        self.sender.take();
+        let result = self
+            .stopped
+            .take()
+            .ok_or(ParentError::Stopped)?
+            .await
+            .map_err(|_| ParentError::Stopped);
+        if let Some(thread) = self.thread.take() {
+            while !thread.is_finished() {
```
erskingardner: Simplification: the `stopped` oneshot plus the 1 ms `is_finished` poll can both go. Have the thread return the ledger (`JoinHandle<JobLedger>`) and join off-reactor:

```rust
pub async fn shutdown(mut self) -> Result<JobLedger, ParentError> {
    self.sender.take();
    let thread = self.thread.take().ok_or(ParentError::Stopped)?;
    tokio::task::spawn_blocking(move || thread.join())
        .await
        .map_err(|_| ParentError::Stopped)?
        .map_err(|_| ParentError::Stopped)
}
```

That removes a field, a channel and a sleep loop, and it still never blocks the reactor, even when dropping the last `Arc<DedupeIndex>` has to wait on RocksDB close.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-24T08:18:12Z` |
| Commit reviewed | `3ae38a1cc97771d994a4d920f6406bb8122485ec` |

> Posted as a comment because GitHub does not allow approving your own PR. The recommendation is still **Merge**.

## Summary

Solid, well-fenced library slice, and it fits the stacked plan (#50 -> #51 -> #52 -> this). The claims I checked hold against the head:

- **Single-attempt fence:** a one-slot `consumed_attempt` is enough because the ledger's `one_active_lease` unique index allows only one active lease at a time. A second attempt can't be served until the first attempt's lease stops being active, so reconnecting to an old attempt is rejected by `verify_active` anyway.
- **Inventory:** it matches the worker's validator. `archived_window` iterates `(timestamp BE, id)` keys inclusively, which fits the worker's strict ordering, window and digest checks. The largest possible 256-item chunk is far below `MAX_FRAME_BYTES`.
- **Authentication and export:** peer UID is checked before the socket reaches the owner thread. Nothing is exported on mismatch, and the lease stays `Leased`.
- **ACKs:** an ACK comes only from `admit_and_accept`. `ProtocolDone` leaves the job in `AwaitingDurability`, and the test proves that completion needs an actual seal.
- **Cancellation:** the `Cancel` guard, the 100 ms poll and the absolute deadline hold together. A queued session that has been cancelled or has expired does not consume the attempt.
- **CI:** all checks pass on this SHA.

Inline notes are all non-blocking: one dead-code cleanup, one duplicate, one error-typing concern worth fixing before the scheduler slice, and a smaller `shutdown`.

One edge case to handle in the scheduler, not here: the async deadline and the inner `until` are the same instant. `serve` can therefore return `TimedOut` after the owner thread has already committed `ProtocolDone` or seen `Failed`. Callers must re-read job state and must not assume the outcome from the error. The ledger fences make this safe, but it should be written down when retry policy lands.

crates/pensieve-ingest/src/sync/parent.rs
```diff
@@ -0,0 +1,838 @@
...
+        };
+        let (reply, response) = oneshot::channel();
+        let command = Command::Session {
+            socket,
+            lease,
+            until,
+            cancelled,
+            reply,
+        };
+        tokio::time::timeout(timeout, self.request(command, response))
```
erskingardner: `tokio::time::timeout` drops this future, including the oneshot `response`, as soon as the deadline fires. The owner thread has no await point the timer can see, and `let _ = reply.send(result)` then discards whatever `exchange` finished.

`upload.receive` can commit `record_protocol_done` (job moves to `AwaitingDurability`) or accept `AttemptFailed` during that window. `serve` still returns `ParentError::Io(TimedOut)`. A caller that treats that error as "session did not finish" will miss a committed protocol completion, and will also lose `Failure::Resource`, which is the only signal this library returns for a split. The receipt stays durable; the `Result` does not describe it.

The capacity-one queue does keep the next `job()` / `lease_next` behind the in-flight command, so rereading the ledger afterward can recover. `serve`'s return value needs to be that outcome. Mark the session cancelled and shut the socket down, but keep awaiting the oneshot. Return `Ok(ProtocolDone)` or `Ok(Failed(...))` when that is what the owner completed. Return the deadline error only when the owner stopped before either outcome.

Please cover this with a regression where a valid `ProtocolDone` commits while the async deadline fires, and assert `serve`'s `Result` matches the ledger.

crates/pensieve-ingest/src/sync/parent.rs
```diff
@@ -0,0 +1,838 @@
...
+    if length == 0 || length > 128 {
+        return Err(ParentError::Protocol(ProtocolError::Limit));
+    }
+    let mut bytes = vec![0; length];
+    socket.read_exact(&mut bytes)?;
+    let hello: Hello = serde_json::from_slice(&bytes).map_err(|_| ProtocolError::Malformed)?;
+    if hello.version != ipc::VERSION {
+        return Err(ParentError::Protocol(ProtocolError::State));
+    }
+    let items = match inventory.archived_window(
```
erskingardner: `DeadlineSocket::read` calls `check()` only before the syscall. A read that starts in time can return successfully up to one 100ms poll later. This path then enters `archived_window` with no further `check()`.

That scan is the heavy operation on the only owner thread: up to `MAX_WINDOW_ITEMS` keys, each probed in the dedupe index, and it does not observe cancellation once started. The session note says a syscall may outlast the deadline by one poll quantum and that no new database operation starts afterward. The upload loop already rechecks after `Frame::read` and before `receive`. The greeting path needs the same recheck, plus `verify_active`, after the hello is parsed and before `archived_window`.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-24T08:20:39Z` |
| Commit reviewed | `3ae38a1cc97771d994a4d920f6406bb8122485ec` |
| Model | `Grok 4.7` |

The inactive single-owner boundary is the right slice: peer UID is checked before any lease export, the constructor requires the writer's own dedupe index, admission happens before ACK, and `ProtocolDone` stays distinct from archive durability.

Two deadline bugs should be fixed before merge. Details are inline. `rustfmt` had passed; clippy, doctests, and the test shards were still pending, so this is from the head commit rather than a finished CI run.

Non-blocking: session tests only cover an empty window. `greeting` accepts `InventoryEnd { count: 0 }` and does not check digest, sequence, or `Assignment::validate`. A fixture with a few archived ids, parsed by `transport::inventory`, would lock chunking, ordering, and the digest the worker actually enforces.

crates/pensieve-ingest/src/sync/parent.rs
```diff
@@ -0,0 +1,980 @@
...
+            delay_terminal_until_cancelled: self.delay_terminal_until_cancelled,
+            socket,
+            lease,
+            until,
+            cancelled,
+            reply,
+        };
+        let request = self.request(command, response);
+        tokio::pin!(request);
+        match tokio::time::timeout(timeout, &mut request).await {
```
erskingardner: **Blocking: this timeout is now redundant, and it reports deadlines as worker EOF.**

`until` is computed just before this timer starts, so the owner's own deadline always expires first, a few microseconds ahead. Every `DeadlineSocket` read and write polls `check()` at least every 100 ms, and a queued command is rejected when `Instant::now() >= until`. Since the `Err` arm now waits on `request` regardless, the only thing this timer adds is waking a blocked syscall up to 100 ms sooner.

The cost of that wake-up is real. `guard.cancel()` calls `shutdown(Both)`, so the owner's blocked `read` returns `Ok(0)`. `DeadlineSocket::read` hands that back without calling `check()`. From there, `read_exact` turns it into `UnexpectedEof`, and `Frame::read` wraps it as `ParentError::Protocol(ProtocolError::Io(UnexpectedEof))`. That is the same value a worker produces when it exits mid-upload. In the greeting phase it becomes `ParentError::Io(UnexpectedEof)`. The async timer fires within microseconds of `until`, while the read only wakes every 100 ms, so a stalled worker will almost always be reported as having hung up rather than `TimedOut`. This is the misclassification the comment above says it prevents. The tests don't catch it because `stalled_greeting…` only asserts `is_err()`.

Suggested fix (net deletion):

```rust
let (reply, response) = oneshot::channel();
self.request(Command::Session { socket, lease, until, cancelled, reply }, response).await
```

Keep `Cancel` as a drop-only guard for external future drops. `Cancel::cancel` can then be folded into `Drop`. The `#[cfg(test)] delay_terminal_until_cancelled` hook (the field, the `Command` field, and the spin loop in the owner) also goes away, because the race it simulates no longer exists. Also assert `ErrorKind::TimedOut` in the stalled-greeting test.

If you want to keep the early wake-up, then at minimum `DeadlineSocket::read` should re-run `self.check()?` when `socket.read` returns `Ok(0)`. That way a self-inflicted shutdown shows up as `ConnectionAborted` or `TimedOut`, not as peer EOF.

crates/pensieve-ingest/src/sync/parent.rs
```diff
@@ -0,0 +1,980 @@
...
+        socket,
+        until,
+        expires_at: lease.expires_at(),
+        cancelled,
+    };
+    socket.check()?;
+    ledger.verify_active(&lease, now())?;
+    if lease.job().since < 0
+        || lease.job().until < lease.job().since
+        || lease.job().until - lease.job().since >= 900
```
erskingardner: Nit: `900` is the same literal as in `Assignment::validate` (`worker/transport.rs`), and `540` at L232 duplicates the worker's `WALL_TIME` (9 × 60). Put these in one shared constant each, next to `MAX_WINDOW_ITEMS`, so the parent and worker limits can't drift apart.

Separately, `JobLedger::enqueue` accepts any `until >= since`. A too-wide job can therefore be leased and then fail here with `InvalidRequest` on every attempt. Enforcing the same constant at enqueue would reject it once, up front. That fix may belong in the scheduler slice.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-24T08:36:22Z` |
| Commit reviewed | `e2793d5aafa7ffdf161d55ac8cb74341cae03f3f` |

> Posted as a comment because GitHub does not allow requesting changes on your own PR. The recommendation is still **Fix blocking issues before merge**.

This is a re-review at `e2793d5`. Most of the earlier feedback is fixed. `RecoveryRequired` and `InvalidRequest` are now separate from worker protocol errors. The greeting path now re-checks `check()`, `verify_active` and `recovery_required` before `archived_window`. A committed `ProtocolDone` is no longer lost when the deadline fires. The 300-ID multi-chunk test now runs through the real worker parser, which covers the non-blocking test-gap note from the earlier Grok review. The premise still holds: this is a small, inactive library slice that fits the stack.

**Blocking (one item, and the fix deletes code):** now that `serve` always awaits the owner's reply, its `tokio::time::timeout` does nothing useful. It also reports a stalled worker's deadline as EOF (`UnexpectedEof`) instead of `TimedOut`. That goes against this revision's stated goal of keeping failure causes typed. I checked the mechanism locally: calling `shutdown(SHUT_RDWR)` on a dup'd socket makes a `recv` blocked on `SO_RCVTIMEO` return `b''` immediately. Details are inline.

**Non-blocking:** the window limit (900) and the nine-minute cap (540) are copied as literals from the worker's own checks. Share the constants instead.

**Still open from the earlier review, not repeated inline:** `DeadlineSocket::check` returns a `Duration` that nobody reads, and it contains a `wall.unwrap()`. `Assignment::for_lease` is called twice. `shutdown` could return the ledger from `JoinHandle` instead of using the `stopped` oneshot plus a 1 ms poll.

**CI:** at the time of review, rustfmt, clippy and doctests pass on this SHA. The three test shards are still running.

crates/pensieve-ingest/src/sync/parent.rs
```diff
@@ -0,0 +1,980 @@
...
+        tokio::pin!(request);
+        match tokio::time::timeout(timeout, &mut request).await {
+            Ok(result) => result,
+            Err(_) => {
+                // Cancelling socket work cannot preempt a SQLite/fsync operation.
+                // Keep the owner reply: a terminal transaction may already have
+                // committed, and must not be misclassified as worker failure.
+                guard.cancel();
+                request.await
+            }
```
erskingardner: The keep-awaiting branch matches the deadline race, and the new test locks a committed `ProtocolDone`. `serve`'s rustdoc still says `timeout` covers queue wait, inventory, and all socket traffic, and that dropping the future does not complete the job. It never says this branch can return `Ok(ProtocolDone)` or `Ok(Failed)` after the timer, or that the second `request.await` runs until the owner finishes a scan or fsync. The session doc already says that. Please put the same two sentences on `serve`, since that is the contract the scheduler will read.

crates/pensieve-ingest/src/sync/parent.rs
```diff
@@ -0,0 +1,980 @@
...
+                            cancelled,
+                            reply,
+                        } => {
+                            let identity = (lease.job().id, lease.job().attempt);
+                            if let Err(error) = ledger.verify_active(&lease, now()) {
+                                let _ = reply.send(Err(error.into()));
+                                continue;
+                            }
+                            if consumed_attempt == Some(identity) {
+                                let _ = reply.send(Err(ProtocolError::State.into()));
```
erskingardner: This rejection happens before any worker byte is read, but it is still `Protocol(State)`. `InvalidRequest` and `RecoveryRequired` now keep the other local faults out of the worker error. A later scheduler that treats `Protocol` as worker misbehavior will classify "this executor already opened the attempt" the same way. A parent-local variant would keep that decision on the caller side.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-24T08:41:18Z` |
| Commit reviewed | `e2793d5aafa7ffdf161d55ac8cb74341cae03f3f` |
| Model | `Grok 4.7` |

> Posted as a comment because GitHub does not allow approving your own PR. The recommendation is still **Merge**.

## Summary

The two deadline bugs from the previous pass are fixed on this head, and CI is green.

- `serve` keeps the owner future when its timer fires, cancels socket work, and still returns a committed `ProtocolDone` or `Failed`. `committed_terminal_outcome_survives_session_deadline` forces that race.
- The greeting path rechecks the deadline, the lease, and recovery before `archived_window`.
- Invalid session parameters and pre-inventory recovery use `InvalidRequest` and `RecoveryRequired`. The 300-id test runs the real worker inventory parser across multiple chunks.

The inactive single-owner boundary holds: peer UID before any lease export, the writer's own dedupe index, admission before ACK, and `ProtocolDone` left in `AwaitingDurability`.

Inline notes are non-blocking. Earlier notes on `DeadlineSocket::check`'s unused duration, the second `Assignment::for_lease`, and the `shutdown` poll are unchanged and still non-blocking.

crates/pensieve-ingest/src/sync/parent.rs
```diff
@@ -0,0 +1,973 @@
...
+    }
+    socket.send(&ParentMessage::InventoryEnd {
+        header: assignment.header(sequence),
+        count: items.len(),
+        digest: hasher.finalize().into(),
+    })?;
+    drop(items);
+    let mut upload = UploadSession::new(lease);
+    loop {
+        let frame = Frame::read(&mut socket)?;
```
erskingardner: Non-blocking, but it matters for the scheduler slice. When `Frame::read` fails on a `DeadlineSocket` error, the error is wrapped as `ProtocolError::Io`. The same parent-owned deadline therefore comes back in different shapes depending on the phase:

- Stalled greeting: `ParentError::Io(TimedOut)`. The new test locks this.
- Stalled upload read: `ParentError::Protocol(ProtocolError::Io(TimedOut))`.
- Stalled ACK write: `ParentError::Io(TimedOut)`, via `send`.
- Peer EOF: `Io(UnexpectedEof)` in the greeting, but `Protocol(Io(UnexpectedEof))` after an ACK. `eof_after_ack_keeps_receipt_for_retry` now locks in that asymmetry.

A scheduler that treats `Protocol` as "the worker misbehaved" would blame the worker for the parent's own deadline in the upload phase. That is the misclassification this revision set out to remove.

The smallest fix is to flatten at this call site:

```rust
let frame = Frame::read(&mut socket).map_err(|error| match error {
    ProtocolError::Io(error) => ParentError::Io(error),
    error => error.into(),
})?;
```

Then update the EOF-after-ACK assertion to `Err(ParentError::Io(e)) if e.kind() == UnexpectedEof`. An alternative is a hand-written `From<ProtocolError>` that always does this, instead of `#[from]`.

crates/pensieve-ingest/src/sync/parent.rs
```diff
@@ -0,0 +1,973 @@
...
+}
+
+struct DeadlineSocket {
+    socket: UnixStream,
+    until: Instant,
+    expires_at: i64,
+    cancelled: Arc<AtomicBool>,
+}
+impl DeadlineSocket {
+    fn check(&self) -> std::io::Result<()> {
```
erskingardner: Nit (simplification): now that nothing reads the duration, both `Option` chains are just comparisons written the long way:

```rust
fn check(&self) -> std::io::Result<()> {
    if self.cancelled.load(Ordering::Acquire) || now() >= self.expires_at {
        return Err(std::io::Error::new(std::io::ErrorKind::ConnectionAborted, "parent session cancelled or expired"));
    }
    if Instant::now() >= self.until {
        return Err(std::io::Error::new(std::io::ErrorKind::TimedOut, "parent session deadline"));
    }
    Ok(())
}
```

A related point: `cancelled` is only set by `Cancel::drop`, and that runs either after the reply arrives or when the `serve` future is dropped. So a caller of `serve` can never see the cancelled branch. Every `ConnectionAborted` a caller does see is a deadline: lease wall-clock expiry here, or `until` having already passed in the queued-command check at L144-148. Consider returning `TimedOut` for those, so that every expiry has one kind.

docs/negentropy_parent_session.md
```diff
@@ -0,0 +1,79 @@
...
+- Create a restricted Unix socket and dedicated worker UID; authenticate the
+  supervised process as well as its UID before interpreting process exit codes.
+- Recover/expire an old active lease before creating a replacement executor after
+  restart. The in-memory one-session fence is not a persistent restart fence.
+- Apply fair, paced retry policy to persisted failure classifications/diagnostics.
+  The library does not automatically split even a Volume report.
+- Keep bounded receipt reconciliation running while admissions are paused. It is
+  available after executor shutdown today; the future long-lived scheduler must
+  add bounded recovery commands/turns instead of opening another ledger.
+- Bind replay source identity to the writer, distinguish sealing-in-progress from
```
erskingardner: Nit: #55 (now integrated) already binds replay's source and dedupe identity to the writer. It also returns a retryable `None` while a seal is in progress, and `negentropy_archived_inventory.md` now documents both. This bullet should list only what is still open, which is the non-destructive guarded cursor repair and disabling the old sync, seed and prune loop.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-24T08:59:40Z` |
| Commit reviewed | `cc4b3e21fe772406d641daf5ec88b697cbf13806` |

> Posted as a comment because GitHub does not allow approving your own PR. The recommendation is still **Merge**.

This is a re-review at `cc4b3e2`. The blocking item from the last pass is fixed, and the fix removes code as suggested:

- `serve` no longer runs its own timer. The owner is now the only thing enforcing the deadline.
- `Cancel` is drop-only.
- The `#[cfg(test)] delay_terminal_until_cancelled` hook is gone.
- The stalled-greeting test now asserts `ErrorKind::TimedOut`.
- A new test checks that a real peer EOF stays `UnexpectedEof`.
- `check()` now returns `()`, so `wall.unwrap()` is gone.
- A reconnect on an already-consumed attempt now returns `InvalidRequest`.
- The `serve` rustdoc now describes when the call can run past its deadline.

The premise still holds: this is a small, inactive library slice that fits the stack.

**Non-blocking (inline):**

- A parent-side deadline still comes back in different shapes depending on the phase. During upload it is wrapped as `Protocol(Io(..))`, and lease expiry shows up as `ConnectionAborted`, not `TimedOut`. There is a small fix that flattens this.
- `check()` can be reduced to two comparisons.
- One "remaining requirement" in the doc is already covered by #55.

**Still open from earlier passes, not repeated inline:**

- `Assignment::for_lease` is still called twice.
- `shutdown` still uses the `stopped` oneshot plus a 1 ms `is_finished` poll, where returning the ledger through `JoinHandle<JobLedger>` would do.
- The `900` and `540` literals still duplicate worker constants.

**CI:** all checks pass on this SHA: rustfmt, clippy, doctests, and all three test shards.

crates/pensieve-ingest/src/sync/parent.rs
```diff
@@ -0,0 +1,973 @@
...
+    })?;
+    drop(items);
+    let mut upload = UploadSession::new(lease);
+    loop {
+        let frame = Frame::read(&mut socket)?;
+        socket.check()?;
+        match upload.receive(ledger, frame, now())? {
+            UploadAction::Event(event) => {
+                socket.check()?;
+                let accepted = upload.admit_and_accept(ledger, event, dedupe, writer, now)?;
```
erskingardner: Blocking: a latched archive on this path is reported as worker protocol failure.

`admit_and_accept` sets the session closed and returns `ProtocolError::State` when `writer.recovery_required()` is set. The `?` here becomes `ParentError::Protocol`, the same variant as a bad sequence or a forged summary. A write that latches recovery inside `write_reserved` comes back as `Protocol(Archive(...))`.

The pre-inventory checks return `ParentError::RecoveryRequired`. Admission is the call that shares the live writer, and `receive` has already committed the receipt before this runs. A caller that treats `Protocol` as worker misbehavior will blame the worker and can retry the relay while the archive latch is set.

If `recovery_required()` is set, return `ParentError::RecoveryRequired` from this arm, including when the latch flips during the write. Add a test that latches the writer and expects that error.

crates/pensieve-ingest/src/sync/parent.rs
```diff
@@ -0,0 +1,973 @@
...
+                            if let Err(error) = ledger.verify_active(&lease, now()) {
+                                let _ = reply.send(Err(error.into()));
+                                continue;
+                            }
+                            if consumed_attempt == Some(identity) {
+                                let _ = reply.send(Err(ParentError::InvalidRequest));
+                                continue;
+                            }
+                            if cancelled.load(Ordering::Acquire) || Instant::now() >= until {
+                                let _ = reply.send(Err(ParentError::Io(std::io::Error::new(
```
erskingardner: Non-blocking: this `until` expiry is `ConnectionAborted` ("cancelled or expired"). `DeadlineSocket::check` reports that same instant as `TimedOut`. A session that expires during `verify_active`, or while the owner is still finishing a dropped call, is therefore indistinguishable from cancellation.

`stalled_greeting_and_dropped_session_does_not_complete_or_orphan_executor` only asserts `TimedOut` for a 30ms socket stall, which holds only when `verify_active` finishes inside that budget. Return `TimedOut` when `Instant::now() >= until`, and keep `ConnectionAborted` for the cancel flag.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-24T09:01:53Z` |
| Commit reviewed | `cc4b3e21fe772406d641daf5ec88b697cbf13806` |
| Model | `Grok 4.7` |

> Posted as a comment because GitHub does not allow requesting changes on your own PR. The recommendation is still **Fix blocking issues before merge**.

The previous deadline bug is fixed on this head. `serve` waits for the owner, a stalled greeting returns `TimedOut`, and a real peer close stays `UnexpectedEof`. CI is green. Peer UID is still checked before any assignment is sent, admission still happens before ACK, and `ProtocolDone` stays in `AwaitingDurability`.

One blocking gap is left: archive recovery discovered during admission is still `ParentError::Protocol(ProtocolError::State)`. Details are inline.

Non-blocking: a deadline that expires before the socket loop is `ConnectionAborted`, the same error as cancellation. Earlier notes on the second `Assignment::for_lease`, the shutdown poll, and the shared `900`/`540` limits still stand.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-24T09:11:05Z` |
| Commit reviewed | `c63355f38fadd09def2b9824b21c01d99ee251f1` |

> Posted as a comment because GitHub does not allow requesting changes on your own PR. The recommendation is still **Fix blocking issues before merge**.

This is a re-review at `c63355f`, covering only the change since `cc4b3e2`.

**What this commit fixes:**

- Recovery latched during admission now returns `RecoveryRequired` before any ACK, in both timings. The test checks that the receipt is kept and no `Accepted` frame is sent. This clears the blocking item from the last Grok pass.
- Queued deadline expiry is now `TimedOut`, and cancellation is `ConnectionAborted`.
- Transport errors during upload are no longer wrapped in `Protocol`, so the greeting phase and the upload phase report EOF and deadlines the same way.

**Blocking (inline):** the new `From<ProtocolError>` impl moves every `ProtocolError::Ledger` into `ParentError::Ledger`. That fixes one misclassification by creating the opposite one. The ledger is where worker input gets checked against durable state. A forged `ProtocolDone` summary, an attempt that goes over `MAX_EVENTS` or `MAX_ATTEMPT_BYTES`, a bad sequence, and an invalid or oversized `AttemptFailed` diagnostic all come back as `LedgerError::Invalid`. Before this commit they were `Protocol(Ledger(Invalid))`. Now they look like a local durable-ledger fault. No parent test sends a bad summary through `serve`, which is why nothing failed.

**Still open from earlier passes, all non-blocking:** `Assignment::for_lease` is still called twice. `shutdown` still uses the oneshot plus a 1 ms poll. The `900` and `540` literals still duplicate worker constants.

**CI:** at the time of review, rustfmt and clippy pass on this SHA. Doctests and the three test shards are still pending.

crates/pensieve-ingest/src/sync/parent.rs
```diff
@@ -0,0 +1,1098 @@
...
+    /// Archive inventory error.
+    #[error(transparent)]
+    Archive(#[from] crate::Error),
+}
+
+impl From<ProtocolError> for ParentError {
+    fn from(error: ProtocolError) -> Self {
+        match error {
+            ProtocolError::Io(error) => Self::Io(error),
+            ProtocolError::Ledger(error) => Self::Ledger(error),
```
erskingardner: **Blocking: this sends worker protocol violations to `Ledger`, the variant for local faults.**

`UploadSession::receive` does not check the summary, the attempt limits or the diagnostic itself. `JobLedger` checks them and returns `LedgerError::Invalid`:

- `record_protocol_done`: `"protocol receipt summary mismatch"` (a forged count or digest) and `"protocol already ended"`.
- `register_received`: `"receipt sequence or attempt limit"` (the worker went over `MAX_EVENTS` or `MAX_ATTEMPT_BYTES`) and `"invalid receipt metadata"`.
- `record_failure_report`: `"invalid failure diagnostic"`, `"failure diagnostic too large"` and `"failure sequence mismatch"`.

Every one of these is caused by the worker. With this impl they come back as `ParentError::Ledger(Invalid(..))`. A scheduler would read that as "local ledger problem, pause" rather than "the worker misbehaved, fail the attempt". That is the reverse of what this commit is meant to do. `Io`, `Archive`, `Budget`, `StaleLease` and `Database` really are local, so flattening them is fine.

Smallest fix: leave `Invalid` inside `Protocol`.

```rust
ProtocolError::Ledger(LedgerError::Invalid(reason)) => {
    Self::Protocol(ProtocolError::Ledger(LedgerError::Invalid(reason)))
}
ProtocolError::Ledger(error) => Self::Ledger(error),
```

Please also add one `serve`-level test that sends `ProtocolDone` with a wrong `count` and expects `Err(ParentError::Protocol(_))`, with the job still `Leased`. Add the same `Invalid` case to `nested_local_upload_faults_do_not_become_worker_protocol_errors`, so the boundary is tested in both directions.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-24T09:14:51Z` |
| Commit reviewed | `c63355f38fadd09def2b9824b21c01d99ee251f1` |
| Model | `Grok 4.7` |

> Posted as a comment because GitHub does not allow requesting changes on your own PR. The recommendation is still **Fix blocking issues before merge**.

The recovery latch and queued-deadline split on this head do what the last review asked. Admission recovery returns `RecoveryRequired` without an ACK, and a queued `until` expiry is `TimedOut`. Peer UID is still checked before any assignment is sent, and `ProtocolDone` still leaves the job in `AwaitingDurability`.

One classification regression should be fixed before merge. `From<ProtocolError>` unwraps every ledger error, so a forged upload summary becomes `ParentError::Ledger`, the local-fault variant. The attempt still fails closed. The type is what the scheduler will branch on.

Non-blocking notes are inline. Lease wall-clock expiry in `DeadlineSocket::check` is still `ConnectionAborted`, the same kind as cancellation. rustfmt, clippy, doctests, and test shards 1 and 3 passed on this SHA; shard 2 was still running.

crates/pensieve-ingest/src/sync/parent.rs
```diff
@@ -0,0 +1,1098 @@
...
+            ProtocolError::Archive(error) => Self::Archive(error),
```
erskingardner: **Blocking: `LedgerError::Invalid` is worker traffic, and this arm reports it as a local ledger fault.**

`receive` maps a forged `ProtocolDone` digest, a bad failure sequence, and an invalid diagnostic to `ProtocolError::Ledger(LedgerError::Invalid(...))`. This `From` then makes that `ParentError::Ledger`. `Budget`, SQLite, and filesystem failures use that same outer variant.

`ParentError::Protocol` is documented as invalid or out-of-order worker traffic. A scheduler that treats `Ledger` as local will not blame the worker for a forged summary, and may treat the ledger as unhealthy. That undoes the split this commit is making.

`forged_summary_and_bad_greeting_fail_closed` only sets the sequence to 9, which fails earlier as `ProtocolError::State`, so it does not catch a correct sequence with a wrong digest.

Keep `Database`, `Io`, `Archive`, and `Budget` as `ParentError::Ledger`. Map `Invalid` back to `ParentError::Protocol`:

```rust
ProtocolError::Ledger(error) => match error {
    error @ LedgerError::Invalid(_) => Self::Protocol(ProtocolError::Ledger(error)),
    error => Self::Ledger(error),
},
```

Add a parent test that sends `ProtocolDone` with the next sequence and a wrong digest, and expects `ParentError::Protocol`.

crates/pensieve-ingest/src/sync/parent.rs
```diff
@@ -0,0 +1,1098 @@
...
+            UploadAction::Event(event) => {
+                socket.check()?;
+                let admission = upload.admit_and_accept(ledger, event, dedupe, writer, now);
+                // Live ingestion can latch the shared writer after inventory,
+                // or this admission itself can latch an uncertain write fault.
+                // The received obligation remains; neither case blames the peer.
+                if writer.recovery_required() {
+                    return Err(ParentError::RecoveryRequired);
+                }
+                let accepted = admission?;
```
erskingardner: **Non-blocking: an unretained live reservation is still blamed on the worker.**

`reserve_unarchived` returns `None` while live ingestion holds the id with `retained == false`. `has_archive_owner` is then false, so `admit_and_accept` returns `ProtocolError::State`. Recovery is not latched, so this `?` becomes `ParentError::Protocol`. That is the shared-dedupe overlap this executor is built to share, not a bad worker frame.

The receipt stays and the job stays `Leased`, but the attempt is already consumed. A scheduler that treats `Protocol` as peer misbehavior will back off the relay. A local retryable error would match the recovery split just above.

erskingardner:
Round-4 correction pushed as d8a9a006b69fc0c437695f052d2ae65d265187f4. Nested LedgerError::Invalid remains a protocol error; actual local storage errors retain their local classification. The real serve regression covers correct-sequence forged count and digest and verifies the job remains leased without ProtocolDone. Documented that ambiguous Protocol(State) is not automatic relay blame. Fresh scoped-clean full just precommit passed with 187 ingester tests. No runtime activation or deployment.

docs/negentropy_parent_session.md
```diff
@@ -0,0 +1,91 @@
...
+Archive recovery latched before or during shared admission returns
+`RecoveryRequired` without an ACK; its received receipt remains durable. Nested
+upload transport/storage/archive failures retain their `Io`/`Ledger`/`Archive`
+classification rather than being wrapped as worker protocol violations. Ledger
+validation of invalid worker traffic (for example a forged completion count or
+digest) remains `Protocol(Ledger(Invalid))`, not a local storage failure. Queued
+deadline expiry returns `TimedOut`, distinct from dropped-session cancellation.
+
+`Protocol(State)` is still ambiguous: an unretained live-ingestion dedupe claim
+or cancellation during admission can produce it without invalid worker traffic.
```
erskingardner: Non-blocking: "cancellation during admission" doesn't match a path I can find. Session cancellation is only checked in `DeadlineSocket::check`, which reports `Io(ConnectionAborted)`, and `admit_and_accept` never reads the cancel flag. What this paragraph leaves out is **lease expiry during admission**. A slow `write_reserved` followed by `accepted` -> `check_ack` with `now >= expires_at` returns `ProtocolError::State`. `receive` has the same check. That is probably the more common innocent cause, so it is the one the scheduler author most needs to see here. Suggest:

```suggestion
`Protocol(State)` is still ambiguous: an unretained live-ingestion dedupe claim
or lease expiry during admission can produce it without invalid worker traffic.
```

crates/pensieve-ingest/src/sync/parent.rs
```diff
@@ -0,0 +1,1152 @@
...
+            ));
+            let ledger = h.executor.shutdown().await.unwrap();
+            let progress = ledger.attempt_progress(h.job, attempt).unwrap();
+            assert_eq!((progress.received, progress.protocol_done), (0, false));
+            assert_eq!(ledger.get(h.job).unwrap().state, JobState::Leased);
+        }
+    }
+
+    #[tokio::test]
+    async fn forged_summary_and_bad_greeting_fail_closed() {
```
erskingardner: Non-blocking: this test's name is now wrong, and it overlaps the new test above. Its "forged summary" case sends `sequence = 9`, and `UploadSession::receive` rejects that as `ProtocolError::State` before the ledger is involved. Nothing in it is a forged summary. The "forged summary" cases are now in `forged_done_summary_is_protocol_error_and_preserves_gap`. Two options:

- Rename this one to `bad_sequence_and_bad_greeting_fail_closed` and tighten its `result.is_err()` to `Protocol(State)` / `Protocol(Limit)`.
- Fold the two new cases into this loop and delete the second harness.

Either way, the test names say what they cover and there is less setup code.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-24T09:23:25Z` |
| Commit reviewed | `d8a9a006b69fc0c437695f052d2ae65d265187f4` |

> Posted as a comment because GitHub does not allow approving your own PR. The recommendation is still **Merge**, once CI is green on this SHA.

This is a re-review at `d8a9a00`, covering only the change since `c63355f`.

**The blocking item from the last pass is fixed.** `LedgerError::Invalid` now stays `Protocol(Ledger(Invalid))`. Other ledger errors (`Database`, `Io`, `Budget`, `StaleLease`) still become `ParentError::Ledger`. I traced every `Invalid` that the upload path can reach:

- `register_received`: the attempt/sequence limits.
- `record_protocol_done`: the summary mismatch.
- `record_failure_report`: an invalid, oversized or out-of-sequence diagnostic.

All of them come from worker input. The only local `Invalid` on that path is `negative clock`, and it cannot happen. So the fix sorts errors the right way and does not create a new misclassification. The new `serve` test drives both forged fields (count and digest) through the real session. It checks that the job stays `Leased`, with no `protocol_done` and no receipts.

**Non-blocking (inline):**

- The new doc paragraph on `Protocol(State)` names the wrong second cause.
- The new test mostly duplicates a neighbouring test whose name is now wrong.

**Non-blocking, for the scheduler slice:** lease expiry still comes back in three shapes, depending on which check notices it first:

- `Io(ConnectionAborted)` from `DeadlineSocket::check`.
- `Protocol(State)` from `receive` or `check_ack`.
- `Ledger(StaleLease)` from `verify_active`.

Now that `Ledger` means a local storage fault, `StaleLease` fits there the least. Mapping `StaleLease` to `Io(TimedOut)` in this `From` impl and in the greeting's `verify_active` calls would give the scheduler one lease-expiry signal. It would also be a small change.

**Still open from earlier passes, all non-blocking:** `Assignment::for_lease` is still called twice. `shutdown` still uses the oneshot plus a 1 ms poll. The `900` and `540` literals still duplicate worker constants.

**CI:** at the time of review, rustfmt passes on this SHA. Clippy, doctests and all three test shards are still running.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-24T09:25:36Z` |
| Commit reviewed | `d8a9a006b69fc0c437695f052d2ae65d265187f4` |
| Model | `Grok 4.7` |

> Posted as a comment because GitHub does not allow approving your own PR. The recommendation is still **Merge**.

## Summary

The blocking classification from `c63355f` is fixed on this head. `LedgerError::Invalid` stays `ParentError::Protocol(ProtocolError::Ledger(Invalid))`, while budget, storage, and archive faults still unwrap to local variants. `forged_done_summary_is_protocol_error_and_preserves_gap` drives both a wrong digest and a wrong count through `serve`, and checks the job stays `Leased` with no receipt and no `ProtocolDone`.

The inactive single-owner boundary still holds: peer UID before any assignment, the writer's own dedupe index, admission before ACK, and `ProtocolDone` left in `AwaitingDurability`. Recovery during admission still returns `RecoveryRequired` without an ACK.

Non-blocking items already on this diff, not repeated inline: the second `Assignment::for_lease`, the `shutdown` oneshot plus 1 ms poll, the duplicated `900`/`540` limits, and lease wall-clock expiry inside `DeadlineSocket::check` still sharing `ConnectionAborted` with cancellation. `Protocol(State)` ambiguity is now documented.

**CI:** rustfmt, clippy, and doctests passed on this SHA. The three test shards were still running.

PR #55, opened 2026-09-24
Fence inventory replay behind checkpointed archive seals

erskingardner:
Stacked on #52. Validates exact writer/dedupe/source identity before cursor initialization and uses a short non-blocking admission snapshot to exclude unfinished seals. Replay never holds the archive gate while streaming; missing files/markers below the completed cutoff remain hard errors. No replay activation, cursor repair, scheduler, or deployment. Full just precommit passed; deterministic rename-before-markArchived race and authority-mismatch regressions included. Independent reviews required before merge.


crates/pensieve-ingest/src/pipeline/segment.rs
```diff
@@ -289,6 +293,47 @@ impl SegmentWriter {
...
+    ) -> Result<Option<u64>> {
+        if !self.uses_dedupe(dedupe) {
+            return Err(Error::Config(
+                "replay dedupe differs from writer authority".to_owned(),
+            ));
+        }
+        if self.recovery_required() {
+            return Err(Error::Config("archive recovery required".to_owned()));
+        }
+        let Some(failure) = self.admission_gate.try_lock() else {
```
erskingardner: Non-blocking suggestion: this ties replay to the admission gate more than it needs to. The gate is also held across the two `canonicalize()` calls below, which are filesystem lookups on the HDD archive. That contradicts the doc's claim that replay "never holds up live archive admission".

A lock-free alternative is to publish the cutoff explicitly:

```rust
/// Exclusive bound of segments whose seal checkpoint fully completed.
checkpointed_before: AtomicU64, // initialized to next_segment
```

Set it at the end of `seal_checkpointed`, after the marker removal and dir sync:

```rust
if let Some(sealed) = &sealed {
    self.checkpointed_before.store(sealed.segment_number + 1, Ordering::SeqCst);
}
```

Then `inventory_source` becomes the identity checks, `recovery_required()`, and a `load()`, returning `Result<u64>`. Several things go away:
- the busy/`Option` path;
- the `failure.is_some()` recheck, which always pairs with `recovery_latched`;
- `uses_dedupe`, which has one caller and can be inlined.

An ordinary in-flight `write` could also no longer turn replay away. The rename-before-marker test still gets `None` through the existing `cursor.next >= ready_before` check.

If you keep the `try_lock` design, at least move the prefix/`canonicalize` identity check above `try_lock()`. It only reads immutable config.

crates/pensieve-ingest/src/sync/inventory.rs
```diff
@@ -130,7 +138,10 @@ impl<'a> InventoryReplay<'a> {
                             )));
                         }
                     }
-                    return Ok(None);
+                    return Err(Error::Validation(format!(
```
erskingardner: With the `cursor.next >= ready_before` guard above, every path through this `NotFound` branch (the `read_dir` loop above included) is now an error:
- a higher segment exists, which gives `missing replay segment N before M`;
- otherwise, `missing checkpointed replay segment N`.

The `read_dir` scan only chooses the message, and it walks the whole archive directory to do it. Suggest deleting the loop and returning one error that includes the cursor and the cutoff. `gap_and_changed_floor_fail_closed_without_advancing` only asserts `is_err()`, so it still covers this case. The doc line on `begin` ("A higher sealed segment with a missing predecessor...") and the "Constant-memory directory discovery" sentence in `docs/negentropy_archived_inventory.md` can be trimmed to match.

crates/pensieve-ingest/src/sync/inventory.rs
```diff
@@ -65,6 +67,9 @@ impl<'a> InventoryReplay<'a> {
         prefix: &str,
         floor: u64,
     ) -> Result<Option<Self>> {
+        let Some(ready_before) = writer.inventory_source(dedupe, archive, prefix)? else {
```
erskingardner: Nit: `inventory_source` already rejects `recovery_required()` and a non-matching canonical archive. That makes the `writer.recovery_required()` term a few lines down redundant, since `step` rechecks it anyway. It also means `begin` canonicalizes the same path a second time. Consider also moving the `cursor.next >= ready_before` early return above the `plain`/`gzip` path construction.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-24T08:19:30Z` |
| Commit reviewed | `4ffdb348d306b94f3775ccb869572d902beba72a` |

The fix is needed and correct as far as I can tell. `segment_number` only moves inside `take_current_for_seal`, which runs under the admission gate for the whole seal (checkpoint, rename, `mark_archived`, marker removal). So a successful `try_lock` snapshot means every segment below the cutoff has been fully checkpointed. A missing file below the cutoff is now a hard error instead of a silent `None`, which is the fail-closed behavior the old docs asked for. Compression is safe here because it renames the gzip before unlinking the plain file. The identity binding (same `Arc<DedupeIndex>`, prefix, canonical dir) is checked before any cursor is initialized, and the tests cover that.

Non-blocking notes (inline):
- **Simpler cutoff:** publishing a `checkpointed_before` atomic at the end of `seal_checkpointed` would let replay skip the admission gate completely. That removes the busy/`Option` path and the `canonicalize()` calls that currently run while the gate is held.
- **Dead directory scan:** now that the cutoff check exists, both outcomes of the `read_dir` gap scan are errors. It only picks the error message, so it can be deleted.
- **Redundant rechecks in `begin`:** these repeat checks that `inventory_source` already did.

Validation gap: rustfmt, clippy and doctests pass. The three test shards were still pending when this was posted.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-24T08:24:27Z` |
| Commit reviewed | `4ffdb348d306b94f3775ccb869572d902beba72a` |
| Model | `Grok 4.7` |

The seal fence is the right fix and it matches the writer’s critical section. `segment_number` advances inside `archive_operation`, and `mark_archived` plus recovery-marker removal both finish before that gate is released. `inventory_source` samples the counter only while holding the same gate, so a successful snapshot is an exclusive cutoff of checkpointed segments; a busy seal returns `None` without cursor or file access. Missing files below that cutoff now fail closed, and archive, prefix, and dedupe identity are checked before any cursor write.

CI is green on this commit. I did not find a blocking correctness, safety, or scope issue. GitHub will not accept an approval from the pull request author, so this merge recommendation is posted as a comment.

crates/pensieve-ingest/src/sync/inventory.rs
```diff
@@ -111,6 +116,9 @@ impl<'a> InventoryReplay<'a> {
         };
         let plain = archive.join(format!("{prefix}-{:09}.notepack", cursor.next));
         let gzip = archive.join(format!("{prefix}-{:09}.notepack.gz", cursor.next));
+        if cursor.next >= ready_before {
```
erskingardner: Non-blocking: this PR gives `begin` a trustworthy cutoff, but first-time initialization (`None => save_cursor(...)` above) still saves any `floor`, including one above it. A mistyped floor (say `100000` for `1000`) gets persisted. From then on every `begin` returns `Ok(None)` here, which looks exactly like "not sealed yet". The doc says a mistyped floor must not be fixed by deleting the database, and no repair path exists yet. So this would be a silent, sticky misconfiguration.

A single guard before `save_cursor` would fail closed:

```rust
None if floor > ready_before => {
    return Err(Error::Config("replay floor is beyond the sealed archive".to_owned()));
}
```

`floor == ready_before` ("start at the next seal") still works. Skip this if a future floor is an intended rollout mode.

crates/pensieve-ingest/src/pipeline/segment.rs
```diff
@@ -983,6 +1033,58 @@ mod tests {
...
+        ];
+        *writer.pause_after_rename.lock() = Some(barriers.clone());
+        let sealing = writer.clone();
+        let handle = std::thread::spawn(move || sealing.seal().unwrap());
+        barriers[0].wait();
+        assert!(archive.join("segment-000000000.notepack").exists());
+        assert!(
+            InventoryReplay::begin(&state, &dedupe, &writer, &archive, "segment", 0)
+                .unwrap()
+                .is_none()
```
erskingardner: Nit: the docs say a busy writer causes no cursor initialization, but this only asserts `None`. The test would still pass if a regression moved `inventory_source` below `save_cursor`. One line pins it:

```rust
assert!(state.replay_cursor().unwrap().is_none());
```

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-24T08:33:49Z` |
| Commit reviewed | `ab6c8842d80e1fe879bbe36ecd4b3335033dc0e4` |

I re-reviewed at `ab6c884`. The only new commit since `4ffdb34` merges master (#52) and doesn't touch the three PR files. The earlier assessment still holds. `segment_number` only advances in `take_current_for_seal`, and every caller of that (`write` → `seal_checkpointed`, `seal`, `Drop`) runs inside `archive_operation`. So a successful `try_lock` snapshot is an exclusive cutoff of fully checkpointed seals. Startup is also safe: the recovery marker is persisted before the rename, and `check_recovery` rejects both the marker and `.open` files, so the startup cutoff can't include a segment whose markers are only half-written. Background compression can't open a window either: it writes `.gz.open`, renames it, and only then unlinks the plain file. Replay tries the plain file first, so it never sees a partial gzip.

The existing inline threads (lock-free `checkpointed_before` atomic, the dead `read_dir` scan, redundant rechecks in `begin`) still apply and I won't repeat them. Two small new notes are inline:
- **Floor validation:** now that the cutoff is trustworthy, a first-time `floor > ready_before` could fail closed instead of persisting a cursor that returns `None` forever.
- **Race test:** it doesn't yet assert that the busy path leaves the cursor untouched, which the docs claim.

Neither blocks merge. GitHub won't accept an approval from the PR author, so this is posted as a comment.

Validation gap: rustfmt, clippy and doctests pass on this SHA. The three test shards were still running when this was posted.

crates/pensieve-ingest/src/pipeline/segment.rs
erskingardner: Non-blocking. This `is_none()` check still passes if `begin` writes the floor cursor and then returns `None`. The busy-path contract is that the cursor stays absent, so assert `state.replay_cursor().unwrap().is_none()` before `barriers[1].wait()`.

If this assert fails, the sealer stays parked on `barriers[1]` and the test hangs until the runner times out. Release that barrier from a drop guard so a regression fails instead of stalling CI.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-24T08:36:15Z` |
| Commit reviewed | `ab6c8842d80e1fe879bbe36ecd4b3335033dc0e4` |
| Model | `Grok 4.7` |

The fence matches the writer critical section. `segment_number` advances only in `take_current_for_seal`, which runs inside `archive_operation` through rename, `mark_archived`, recovery-marker removal, and the directory sync. `inventory_source` reads that counter only while it holds the same gate, so a successful snapshot is an exclusive cutoff of checkpointed segments. A busy gate returns `None` before any cursor or file access. Missing files below the cutoff fail closed. Archive, prefix, and the exact dedupe object are rejected before any cursor write. Compression renames the gzip into place before unlinking the plain file, so the plain-then-gzip open cannot observe a hole.

The optional notes already on the PR (an atomic cutoff, dropping the error-only directory scan, and the repeated checks in `begin`) are simplifications. They can wait. One test gap is inline.

rustfmt, clippy, and doctests passed on this commit. The three test shards were still running when this was posted. GitHub will not accept an approval from the pull request author, so this merge recommendation is a comment.

PR #56, opened 2026-09-24
Add bounded parent maintenance and uncertain-outcome recovery

erskingardner:
## Scope

Stacked on #54. Adds library-only commands on the existing single owned executor: expire, retry, retry_durability, bounded receipt maintenance, and single-attempt progress/failure-report reads. No scheduler, listener, process launch, replay activation, or production changes.

## Safety and bounds

- Persisted keyset scan visits at most 32 jobs and 256 total receipts per maintenance turn, including retry-wait, blocked, split, and zero-receipt awaiting jobs.
- Reconciliation commits before the job cursor advances; errors preserve the failing position and crashes may safely replay work.
- Recovery bypasses admission ceilings but not real disk errors. Failed attempts without ProtocolDone never become complete.
- Prototype schema 5 adds the cursor; old prototypes fail closed without mutation or automatic migration. Failure reports and receipt obligations are retained.
- Cancellation does not roll back committed owner commands; bounded reads support uncertain-outcome recovery.

## Verification

Full just precommit passed immediately before commit: workspace formatting, strict Clippy, all workspace tests and doctests. The ingester library passed 187 tests. Regression coverage includes fairness, reopen, quota exhaustion, stale retry, dropped responses, zero-event durability, total receipt bounds, cursor write faults, and failure-report to split/blocked to maintenance to reopen preservation without false completion.

## Remaining gates

Exact-head CI and independent review remain required. Scheduler activation must address maintenance cadence through retained history and stale caller timestamps behind queued owner work; both limitations are explicitly documented. No reviewers have been launched for this PR yet.


crates/pensieve-ingest/src/sync/jobs.rs
```diff
@@ -677,6 +680,76 @@ impl JobLedger {
...
+        }
+        let after: i64 = self.db.query_row(
+            "SELECT scan_job FROM receipt_totals WHERE singleton=1",
+            [],
+            |r| r.get(0),
+        )?;
+        let jobs = {
+            let mut query = self
+                .db
+                .prepare("SELECT id,state FROM jobs WHERE id>?1 ORDER BY id LIMIT ?2")?;
```
erskingardner: The only thing that forces a full-history walk here is the missing index. The ledger is being bumped to schema 5 anyway and is unshipped, so a partial index is free right now:

```sql
CREATE INDEX unresolved_jobs ON jobs(id)
  WHERE state IN ('awaiting_durability','retry_wait','blocked','split');
```

Then the query becomes `SELECT id FROM jobs WHERE id>?1 AND state IN (...same list...) ORDER BY id LIMIT ?2`. Each turn visits at most 32 *unresolved* jobs and never touches complete or queued history. The Rust state check goes away, and so does the "100,000 jobs require about 3,126 turns" limitation in `negentropy_parent_maintenance.md`. Leased jobs drop out naturally. The in-transaction `Leased` check in `reconcile_archived` still guards the edge case. If you defer this, it will need a schema-6 bump later.

crates/pensieve-ingest/src/sync/jobs.rs
```diff
@@ -677,6 +680,76 @@ impl JobLedger {
...
+            self.db
+                .execute("UPDATE receipt_totals SET scan_job=0 WHERE singleton=1", [])?;
+            progress.wrapped = true;
+            return Ok(progress);
+        }
+        for (id, state) in jobs {
+            if progress.checked == max_receipts {
+                break;
+            }
+            if state != "leased" && state != "complete" {
```
erskingardner: `queued` jobs pass this filter, but they are always a no-op. `enqueue` and `split` only ever insert queued rows with no receipts, and `complete_if_durable` only completes `awaiting_durability`/`split` rows. Each one still costs an `IMMEDIATE` transaction with three writes and a `synchronous=FULL` commit. A newly enqueued sweep therefore burns entire turns on fsyncs that change nothing. Suggest `matches!(state.as_str(), "awaiting_durability" | "retry_wait" | "blocked" | "split")`, or make this unnecessary with the partial index above.

crates/pensieve-ingest/src/sync/jobs.rs
```diff
@@ -677,6 +680,76 @@ impl JobLedger {
...
+                let outcome =
+                    self.reconcile_archived(id, dedupe, writer, max_receipts - progress.checked)?;
+                progress.checked += outcome.checked;
+                progress.satisfied += outcome.satisfied;
+                progress.completed += u32::from(outcome.complete);
+            }
+            // Separate commits deliberately favor harmless replay over skipping.
```
erskingardner: Right now each visited job gets its own durable commit, including skipped complete/leased rows. That's up to 32 extra fsyncs per turn just to move a counter. The doc already says replay is harmless, so writing the cursor once per turn is just as safe. Track `last = id` in the loop, and persist it once at the end and on the error path, e.g. by wrapping the loop body in a closure. A crash then replays at most one turn, which is the same guarantee with far fewer commits.

crates/pensieve-ingest/src/sync/jobs.rs
```diff
@@ -677,6 +680,76 @@ impl JobLedger {
...
+                .query_map(params![after, max_jobs], |r| {
+                    Ok((r.get::<_, i64>(0)?, r.get::<_, String>(1)?))
+                })?
+                .collect::<Result<Vec<_>, _>>()?
+        };
+        let mut progress = MaintenanceProgress::default();
+        if jobs.is_empty() {
+            self.db
+                .execute("UPDATE receipt_totals SET scan_job=0 WHERE singleton=1", [])?;
+            progress.wrapped = true;
```
erskingardner: Small simplification: if the query returned fewer than `max_jobs` rows, the end of the keyspace has already been reached. Reset `scan_job` to 0 and set `wrapped` in that same turn, and drop the separate empty-result branch. As written, every full pass costs one extra call that does no work.

docs/negentropy_process_isolation_plan.md
```diff
@@ -0,0 +1,269 @@
+# Negentropy process isolation: implementation proposal
```
erskingardner: This is a 269-line historical proposal that has nothing to do with parent maintenance. Consider landing it (and the execution-plan cross-link) as a separate docs-only PR so this diff stays about the maintenance mechanism.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-24T09:00:37Z` |
| Commit reviewed | `5bd975382ee8c0a75429fd6183b1bc93f2e06ca9` |

I found no correctness bugs. The mechanism is right: the per-job `(scan_attempt, scan_sequence)` cursor plus the new cross-job `scan_job` cursor means one stuck receipt or one large job can't starve recovery. Leased jobs are rejected inside the reconcile transaction. Failed attempts can't be promoted, because `complete_if_durable` still requires `protocol_done=1 AND archived=received`. The new owner commands are thin pass-throughs and match the existing `Lease`/`Job` pattern. The tests are targeted and check what the PR claims: fairness, reopen, admission ceiling bypass, cursor fault, dropped reply, and stale-attempt fencing.

Nothing here blocks merge. The notes below are all efficiency, simplification or scope:

1. **Unresolved-job partial index (worth doing now).** The schema is already moving to v5 and is unshipped, so adding `CREATE INDEX unresolved_jobs ON jobs(id) WHERE state IN ('awaiting_durability','retry_wait','blocked','split')` now costs nothing. Later it would cost a v6 bump. It makes maintenance cost scale with outstanding work instead of with total history. That removes the documented cadence limitation (about 3,126 turns per 100k jobs) and the Rust-side state filter.
2. **No-op write transactions.** Queued jobs go through a full `IMMEDIATE` reconcile transaction. Queued jobs never have receipts and can never be completed by `complete_if_durable`. Every visited job also gets its own cursor commit. With `synchronous=FULL`, a turn over 32 queued or complete jobs costs up to 64 fsyncs and does no work. Skip `queued`, and persist the cursor once per turn.
3. **Extra empty wrap turn.** When fewer than `max_jobs` rows come back, you already know you've reached the end, so the cursor can reset in the same turn.
4. **Scope.** The 269-line verbatim historical proposal doc and the stack status table aren't part of maintenance. They'd be cleaner as a separate docs-only PR. The "Under review" rows will also go stale as each PR in the stack merges.

Minor chaff: the existing `AttemptFailed` test changes `FailureKind::Relay` to `Volume` and `Vec::new()` to `vec![]`. The new assertions work with either value, so these edits aren't needed.

**Validation gap:** At review time, rustfmt, clippy and doctests are green on this head, but the three test shards are still pending. Merge only after they pass, and after #54, the base, lands. (GitHub doesn't allow approving your own PR, so this is posted as a comment.)

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-24T09:04:19Z` |
| Commit reviewed | `5bd975382ee8c0a75429fd6183b1bc93f2e06ca9` |
| Model | `Grok 4.7` |

This is the right next library slice. Expire, retry, durability retry, bounded receipt maintenance, and single-attempt progress/failure reads all sit on the existing single owner. Nothing here starts a scheduler, listener, or production path.

The maintenance turn stays inside the stated bounds: at most 32 jobs and 256 receipts total, live and completed jobs are skipped, and the persisted job cursor advances only after that job's reconcile commits. A cursor-write failure leaves the failed job selected. A crash between the two commits only repeats work that is safe to repeat. Receipt compaction still cannot turn a failed attempt into success without ProtocolDone. Schema 5 is create-only; older prototypes fail closed with no migration. Recovery skips admission ceilings and still returns real disk errors. The new parent methods are the same capacity-one owner queue, so a maintenance turn cannot overlap an upload session.

The new tests cover cross-job fairness, reopen, failure-report preservation through split and blocked, the total receipt cap, a failed cursor write, stale durability retry, and a dropped terminal response. On this commit, clippy, rustfmt, doctests, and test shards 2 and 3 passed. Test shard 1 was still running at review time; let it finish before merge.

The execution checklist still lists #53 as under review. That PR merged after this commit. Correct that row the next time the checklist is edited.

GitHub rejected an approve event because this review is from the pull request author. The recommendation above is still to merge.

erskingardner:
Integrated parent-session classification correction d8a9a00 without changing maintenance behavior. New head 782c47cc26f3058a851e63c5b20009adcc814824. Fresh scoped-clean full just precommit passed with 195 ingester tests, including correct-sequence forged-summary rejection and maintenance cursor/fairness/failure-report preservation regressions. No scheduler activation or deployment.

crates/pensieve-ingest/src/sync/jobs.rs
```diff
@@ -677,6 +682,81 @@ impl JobLedger {
...
+        }
+        let after: i64 = self.db.query_row(
+            "SELECT scan_job FROM receipt_totals WHERE singleton=1",
+            [],
+            |r| r.get(0),
+        )?;
+        let jobs = {
+            let mut query = self
+                .db
+                .prepare("SELECT id FROM jobs WHERE id>?1 AND state IN ('awaiting_durability','retry_wait','blocked','split') ORDER BY id LIMIT ?2")?;
```
erskingardner: SQLite doesn't pick `unresolved_jobs` for this query. The ledger never runs `ANALYZE`, so the planner goes by its default heuristics and chooses the `due_jobs` index plus a sort. I checked on SQLite 3.51 and 3.53, using this exact schema and query, with and without 100k rows of history:

```
SEARCH jobs USING COVERING INDEX due_jobs (state=?)
USE TEMP B-TREE FOR ORDER BY
```

So every turn reads and sorts all unresolved rows after the cursor, and the new index only adds write cost. With `INDEXED BY` the plan becomes `SEARCH jobs USING INDEX unresolved_jobs (id>?)`, a real O(max_jobs) keyset. It also makes the query fail loudly if the index ever goes missing:

```suggestion
                .prepare("SELECT id FROM jobs INDEXED BY unresolved_jobs WHERE id>?1 AND state IN ('awaiting_durability','retry_wait','blocked','split') ORDER BY id LIMIT ?2")?;
```

The other option is to drop the index and rely on `due_jobs`, whose cost is also bounded by unresolved rows, and change the doc to match. Either way, the doc sentence about a "partial-index keyset" is currently not true. I didn't check the bundled SQLite 3.46 (libsqlite3-sys 0.30.1). An `EXPLAIN QUERY PLAN` assertion in a test would settle that.

crates/pensieve-ingest/src/sync/jobs.rs
```diff
@@ -712,6 +792,21 @@ pub struct ReceiptReconciliation {
     pub complete: bool,
 }
 
+/// Bounded maintenance accounting; zero changes is not a stop condition.
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
+pub struct MaintenanceProgress {
+    /// Job rows handled this turn, including skipped live/completed rows.
```
erskingardner: This doc is out of date. After `3020f7f` the query never selects live or completed rows, so nothing gets skipped. Suggest: `/// Unresolved job rows reconciled this turn.`

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-24T09:26:29Z` |
| Commit reviewed | `782c47cc26f3058a851e63c5b20009adcc814824` |

This is a follow-up to the two reviews of `5bd9753`. Commit `3020f7f` handles most of the earlier notes. Queued, complete and leased rows no longer reach a reconcile transaction. The cursor is written once per turn, and the scan wraps in the same turn once it knows it has reached the end. The new tests pin down each of these: no writes over empty history, one cursor write before an error, and a wrap on the last turn. I found no correctness bugs in `maintain_archived` or in the new owner commands. The error cursor, the receipt budget across jobs, the wrap condition, and the rule that a failed attempt never completes all behave as described.

There is one real gap, and it is non-blocking. **SQLite does not use the new `unresolved_jobs` partial index.** The ledger never runs `ANALYZE`, so the planner chooses `due_jobs (state=?)` and then sorts every unresolved row after the cursor into a temp B-tree. So per-turn cost still grows with unresolved work, not with history, but the new index is only extra write cost, and the doc's "partial-index keyset" claim is wrong. The fix is one token (see the inline comment), and it needs no schema bump.

Minor:
- The `MaintenanceProgress::jobs` doc is out of date (inline comment).
- The execution checklist lists #57 as "Review/CI gate", but #57 has merged.
- The test chaff from the earlier review is still in the diff (`Relay` changed to `Volume`, `Vec::new()` changed to `vec![]`, `&path` changed to `path`). None of it is needed.
- The 269-line historical proposal doc is still out of scope for a maintenance PR. This was raised before, and the author kept it on purpose, so I'm not raising it again.

**Validation gap:** at review time, rustfmt, clippy and doctests pass on this head. All three test shards are still pending. Merge only after they pass and after #54 (the base) lands. GitHub rejects an approve from the PR author, so this review is posted as a comment.

crates/pensieve-ingest/src/sync/jobs.rs
```diff
@@ -613,6 +682,81 @@ impl JobLedger {
...
+        max_receipts: u32,
+    ) -> Result<MaintenanceProgress, LedgerError> {
+        if max_jobs == 0
+            || max_jobs > MAX_RECOVERY_JOBS
+            || max_receipts == 0
+            || max_receipts > MAX_RECEIPT_BATCH
+            || writer.recovery_required()
+            || !writer.uses_dedupe(dedupe)
+        {
+            return Err(LedgerError::Invalid(
```
erskingardner: `recovery_required` is bundled into `LedgerError::Invalid` together with a zero bound and a dedupe mismatch. `serve` reports that same latch as `ParentError::RecoveryRequired`. This command is what a later scheduler will poll between sessions, and `ParentError::Ledger(Invalid(_))` will not select that pause.

Check the latch before the bounds test and return `RecoveryRequired` from the owner arm. Leave `Invalid` for bad limits and a mismatched dedupe. The mid-turn refusal in `reconcile_archived` still surfaces as `Invalid("archive recovery required")` and needs the same mapping on this command.

crates/pensieve-ingest/src/sync/jobs.rs
```diff
@@ -648,6 +792,21 @@ pub struct ReceiptReconciliation {
     pub complete: bool,
 }
 
+/// Bounded maintenance accounting; zero changes is not a stop condition.
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
+pub struct MaintenanceProgress {
+    /// Job rows handled this turn, including skipped live/completed rows.
```
erskingardner: `jobs` counts rows actually passed to `reconcile_archived`. Queued, complete, and leased rows never enter the partial index, so they are not included. An empty unresolved set returns `jobs == 0` with `wrapped == true`.

Describe this as reconciled unresolved jobs.

docs/negentropy_execution_plan.md
```diff
@@ -34,8 +40,13 @@ analytics architecture. The existing SDK remains pinned; no SDK fork is required
 | 2b, #49 | Independent periodic archive sealing | Merged; Parquet-disabled durability and joined shutdown tests |
 | 2c, #50 | Bounded sealed-segment inventory replay/export | Merged; no runtime replay enabled |
 | 3a, #51 | One-job worker executable and parent wire exchange | Merged; CI and both independent reviews passed; no production parent endpoint |
-| 3b prerequisite | Distinct worker volume/event-size outcomes | Implemented; test/review gate, no scheduler activation |
-| 3b | Parent scheduler, authenticated listener and inventory activation | Next; persistence/fairness, bounded executor, seal-race/cursor repair gates |
+| 3b prerequisite, #52 | Distinct worker volume/event-size outcomes and request-scoped capture | Merged; no scheduler activation |
+| 3b prerequisite, #57 | Explicit inventory cursor rewind | Review/CI gate; no operator CLI or activation |
```
erskingardner: #57 merged at 2026-09-24T09:11:30Z, before this commit. This table is the status authority, and the neighboring rows were updated in this PR. Mark this cell Merged and keep the "no operator CLI or activation" clause.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-24T09:30:43Z` |
| Commit reviewed | `782c47cc26f3058a851e63c5b20009adcc814824` |
| Model | `Grok 4.7` |

This is the right next library slice. Expire, same-window retry, durability retry, bounded receipt maintenance, and single-attempt progress and failure-report reads all sit on the existing single owner. Nothing here starts a scheduler, listener, or production path.

The maintenance turn stays inside the stated bounds. The partial index selects only unresolved jobs, at most 32 per turn and 256 receipts total. The cross-job cursor advances only through jobs whose reconcile committed, and it is written once per turn, including when a later job fails. A failed cursor write leaves the previous position. A crash between the reconcile commit and that write repeats at most one bounded turn. `complete_if_durable` still requires the current attempt's `protocol_done` and an empty receipt set, so compacting a failed attempt cannot mark it complete. Schema 5 is create-only; older prototypes fail closed with no migration. Retry, expire, and maintenance skip admission ceilings and still surface real disk errors.

The new tests cover cross-job fairness, reopen while admission is paused, failure-report preservation through split and blocked, the shared receipt cap, a rejected cursor write, success committed before a later fault, same-turn wrap, stale durability retry, and a dropped terminal response. On this head, rustfmt, clippy, doctests, and all three test shards passed.

The inline notes are follow-ups. None of them block merge.

GitHub rejects an approve event from the pull request author. The recommendation above is still to merge.

erskingardner:
Addressed the latest maintenance findings in 4b42409f6f99a81a0ed032f5c563d803277b0368 and integrated master aa7ee02, retaining all inventory-rewind changes. Maintenance explicitly selects unresolved_jobs; the bundled SQLite EXPLAIN regression proves indexed id keyset access without a temporary sort. Owner maintenance maps preexisting and mid-turn archive recovery to RecoveryRequired, with deterministic durable-obligation/reopen regression. Updated job-count documentation, merged prerequisite statuses, and lease-expiry ambiguity wording. Fresh scoped-clean full just precommit passed: 202 ingester tests. No activation or deployment.

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-24T09:40:35Z` |
| Commit reviewed | `4b42409f6f99a81a0ed032f5c563d803277b0368` |

This is a follow-up to the reviews of `782c47c`. Commit `4b42409` fixes the issues found in that round. The maintenance query now uses `INDEXED BY unresolved_jobs`, and `maintenance_query_uses_partial_index_without_sorting_history` pins the plan on the bundled SQLite. `MaintenanceProgress::jobs` now has an accurate doc. The execution checklist marks #52/#53/#54/#55/#57 as merged. The owner's `Maintain` arm reports the recovery latch as `ParentError::RecoveryRequired`, both before the turn and when the latch is raised during it. The second case is tested deterministically without a production hook.

I re-checked `maintain_archived` at this head and found no correctness bugs:
- The receipt budget is shared across jobs.
- `last` only moves past jobs whose reconcile committed.
- The scan wraps only when the whole short page was handled.
- Parents always have smaller ids than their split children, so completion propagated in the same turn is never counted twice.
- `complete_if_durable` still requires `protocol_done=1 AND archived=received` for the current attempt.

The new owner commands are thin pass-throughs on the same capacity-one queue. Their tests cover stale-lease fencing, a dropped reply, and reading state back after a lost terminal response.

Nothing blocks merge. The inline notes are nits:
- Two new `ParentExecutor` methods have no blank line before the next method.
- One test makes three identical empty-history calls where one is enough under the current wrap logic.

The test-fixture chaff and the out-of-scope historical proposal doc were raised in earlier rounds, and the author kept them on purpose, so I'm not raising them again.

**Validation gap:** at review time, rustfmt, clippy and doctests pass on this head. All three test shards are still running. Merge only after they pass. I did not run anything locally. GitHub rejects an approve from the PR author, so this review is posted as a comment.

crates/pensieve-ingest/src/sync/parent.rs
```diff
@@ -215,12 +302,120 @@ impl ParentExecutor {
...
+        self.request(
+            Command::Maintain {
+                jobs,
+                receipts,
+                reply,
+            },
+            response,
+        )
+        .await
+    }
```
erskingardner: Nit: this method runs straight into the next doc comment with no blank line. Every other method in this impl has one. rustfmt keeps whatever is there, so CI won't catch it.

```suggestion
    }

```

crates/pensieve-ingest/src/sync/parent.rs
```diff
@@ -215,12 +302,120 @@ impl ParentExecutor {
...
+            Command::AttemptProgress {
+                job,
+                attempt,
+                reply,
+            },
+            response,
+        )
+        .await
+    }
     /// Authenticate before exporting a lease, then perform exactly one exchange.
```
erskingardner: Nit: this method runs straight into the next doc comment with no blank line. Every other method in this impl has one. rustfmt keeps whatever is there, so CI won't catch it.

```suggestion
    }

```

crates/pensieve-ingest/src/sync/jobs.rs
```diff
@@ -1049,6 +1144,260 @@ mod tests {
...
+            db.maintain_archived(&dedupe, &writer, 32, 256)
+                .unwrap()
+                .jobs,
+            0
+        );
+        assert!(
+            db.maintain_archived(&dedupe, &writer, 32, 256)
+                .unwrap()
+                .wrapped
+        );
```
erskingardner: Nit: since `3020f7f`, an empty selection resets the cursor and sets `wrapped` in the same turn. All three calls here therefore do the same thing: `jobs == 0`, `wrapped == true`, no write. The repeats come from the old multi-turn wrap. One call that checks both fields proves the same thing:

```suggestion
        let progress = db.maintain_archived(&dedupe, &writer, 32, 256).unwrap();
        assert_eq!((progress.jobs, progress.wrapped), (0, true));
```

erskingardner (commented):
| Review metadata | Value |
| --- | --- |
| Reviewed at (UTC) | `2026-09-24T09:44:06Z` |
| Commit reviewed | `4b42409f6f99a81a0ed032f5c563d803277b0368` |
| Model | `Grok 4.7` |

This head fixes the follow-ups from `782c47c`. `INDEXED BY unresolved_jobs` is the maintenance query, and the bundled `EXPLAIN QUERY PLAN` test requires an `id>?` search with no temp sort. The owner maps a recovery latch seen before the turn and one raised during it to `RecoveryRequired`. `reconcile_archived` still rolls back the open job if that latch is set before commit, and `complete_if_durable` still requires `protocol_done` plus an empty receipt set.

The turn bounds hold. At most 32 unresolved jobs and 256 receipts are visited. The cross-job cursor is written once, only through jobs whose reconcile committed, and a short final page wraps in that same turn. A failed job stays selected. Schema 5 is create-only; older prototypes still fail closed with no migration. Expire, retry, and durability retry skip admission ceilings and still return real disk errors.

Rustfmt, clippy, doctests, and all three test shards passed on this commit. #54, the base, has merged. Nothing here starts a scheduler, listener, or production path.

GitHub rejects an approve event from the pull request author. The recommendation above is still to merge.

